PR #1, opened 2023-03-08
Initialize tauri mobile app (svelte frontend)

daniel-mader:


daniel-mader:
Please build this branch with all your available targets (Desktop, Android, iOS) to verify it works on your machine as well

PR #5, opened 2023-03-16
feat: typescript bindings

daniel-mader:


src/stores.ts
```diff
@@ -0,0 +1,11 @@
+import { readable } from 'svelte/store';
+import { listen } from '@tauri-apps/api/event';
+import type { TransferState as State } from '../src-tauri/bindings/TransferState'; // TODO: copy to /src instead of importing across root
+
+export const state = readable<State>(undefined, (set) => {
+  listen('state-changed', (event) => {
+    console.log(event);
+    set(event.payload as State);
+  });
+  // TODO: unsubscribe from listener?
```
nanderstabel, line 10: What would it mean to unsubscribe from the listener?

src-tauri/src/command.rs
```diff
@@ -17,10 +19,19 @@ pub struct CommandMessage {
...
+    info!("received command `{}` with payload `{}`", command, payload);
 
     match command.as_str() {
+        "[INIT] Get initial state" => {
+            let initial_state = AppState::new(StateStatus::Stable);
+            let transfer_state = TransferState::from(initial_state);
+            window.emit("state-changed", transfer_state).unwrap();
+            // TODO: should not be necessary to create new, allow Copy trait on TransferState?
+            info!("emitted event `{}` with payload `{:?}`", "state-changed", TransferState::from(AppState::new(StateStatus::Stable)));
+            Ok(TransferState::from(AppState::new(StateStatus::Stable)))
```
nanderstabel: I think passing the value by reference instead of copying the state is sufficient in this case, e.g.:
```suggestion
            let transfer_state = TransferState::from(AppState::new(StateStatus::Stable));
            info!(
                "emitted event `{}` with payload `{:?}`",
                "state-changed", &transfer_state
            );
            Ok(transfer_state)
```

nanderstabel (approved):
Looks good!

src/stores.ts
daniel-mader: I will have to check at a later point if leaving the subscriber active is a problem or not

PR #6, opened 2023-04-18
feat: welcome flow, redux pattern for state management, i18n, dev mode

daniel-mader:
# Description of change
First meaningful app flow. User starts from a blank app, gets prompted for a name and a DID is created in the background, then the user is directed to the "Profile" page.

- n/a

- Frontend integration tests, backend unit tests (where applicable), manual exploration tests


src-tauri/src/state/persistence.rs
```diff
@@ -0,0 +1,50 @@
...
+use tracing::info;
+// use tauri::path::resolve_path;
+
+use crate::state::state::TransferState;
+
+/// Loads a [TransferState] from the app's data directory.
+/// If it does not exist or it cannot be parsed, it will fallback to default values.
+pub async fn load_state(_app_handle: tauri::AppHandle) -> anyhow::Result<TransferState> {
+    //TODO: use tauri::path::BaseDirectory::AppData;
+    let state_file_path = dirs_next::data_dir().unwrap().join("com.tauri.dev").join("state.json");
```
nanderstabel: This line (idem for L25 and L36) results in a panic for me:
```
04-20 12:40:18.229 32640 32674 I RustStdoutStderr: 2023-04-20T10:40:18.229031Z  INFO identity_wallet::command: received action `GetState` with payload `None`
04-20 12:40:18.229 32640 32674 I RustStdoutStderr: thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', src/state/persistence.rs:14:49
```

Turns out that dirs_next does not support Android. Appearantly this is/was also an issue for Tauri:
https://github.com/tauri-apps/tauri/issues/6054#issuecomment-1382822014.

I found [another crate](https://docs.rs/app_dirs2/latest/app_dirs2/) that does work for android, but I don't know if it
works for other platforms as well. So I would suggest something like this:
```
    let state_file_path = if cfg!(android) {
        app_dirs2::data_root(AppDataType::UserData)?
            .join("com.tauri.dev")
            .join("state.json")
    } else {
        dirs_next::data_dir()
            .ok_or(anyhow::anyhow!("Data directory not found"))?
            .join("com.tauri.dev")
            .join("state.json")
    };
```
I've also replaced the `unwrap()`. In most cases you'd want to avoid using `unwrap()`, in this case I would suggest to
use `Option::ok_or`, which maps an `Option` to a `Result`. With the use of `anyhow!` you can conveniently add the error message.

src-tauri/src/state/persistence.rs
```diff
@@ -0,0 +1,50 @@
...
+    let bytes = read(&state_file_path).await?;
+    let content = String::from_utf8(bytes)?;
+    let transfer_state: TransferState = serde_json::from_str(&content).unwrap();
```
nanderstabel:
```suggestion
    let transfer_state: TransferState = serde_json::from_str(&content)?;
```

src-tauri/src/state/persistence.rs
```diff
@@ -0,0 +1,50 @@
...
+        .await?;
+    info!("state saved to app_data_dir: {}", state_file_path.display());
+    Ok(())
+}
+
+// Removes the state file from the app's data directory.
+pub async fn delete_state(_app_handle: tauri::AppHandle) -> anyhow::Result<()> {
+    //TODO: use tauri::path::BaseDirectory::AppData;
+    let state_file_path = dirs_next::data_dir().unwrap().join("com.tauri.dev").join("state.json");
+    remove_file(&state_file_path).await.unwrap();
```
nanderstabel:
```suggestion
    remove_file(&state_file_path).await?;
```

src-tauri/src/state/persistence.rs
```diff
@@ -0,0 +1,50 @@
...
+    info!("state loaded from app_data_dir: {}", state_file_path.display());
+    Ok(transfer_state)
+}
+
+/// Persists a [TransferState] to the app's data directory.
+pub async fn save_state(_app_handle: tauri::AppHandle, transfer_state: TransferState) -> anyhow::Result<()> {
+    //TODO: use tauri::path::BaseDirectory::AppData;
+    let state_file_path = dirs_next::data_dir().unwrap().join("com.tauri.dev").join("state.json");
+    let mut file = File::create(&state_file_path).await?;
+    file.write_all(serde_json::to_string(&transfer_state).unwrap().as_bytes())
```
nanderstabel:
```suggestion
    file.write_all(serde_json::to_string(&transfer_state)?.as_bytes())
```

src-tauri/src/state/actions.rs
```diff
@@ -0,0 +1,27 @@
...
+pub enum ActionType {
+    #[serde(rename = "[App] Get state")]
+    GetState,
+    #[serde(rename = "[App] Reset")]
+    Reset,
+    #[serde(rename = "[DID] Create new")]
+    CreateNew,
+    #[serde(rename = "[Settings] Set locale")]
+    SetLocale,
+}
```
nanderstabel, line 27: This list will eventually quite big, so at some point we should adjust the (de)serialization so that the variants are mapped a bit more layered somehow. Don't know yet how to do that exactly (and how that will work with TS bindings), but it could look something like this pseudocode:
```rust
#[ts(export)]
pub enum ActionType {
    #[serde(rename = "[App] {action}")]
    App(App),
    #[serde(rename = "[DID] {action}")]
    DID(DID),
    #[serde(rename = "[Settings] {action}")]
    Settings(Settings),
}

pub enum App {
    #[serde(rename = "Get state")]
    GetState,
    #[serde(rename = "Reset")]
    Reset

}

...etc

```

src-tauri/src/command.rs
```diff
@@ -1,39 +1,61 @@
...
+                .await
+                .unwrap();
+        }
+        ActionType::SetLocale => {
+            set_locale(app_state.inner(), Action { r#type, payload }).unwrap();
+            save_state(app_handle, TransferState::from(app_state.inner()))
+                .await
+                .unwrap();
+        }
+    };
```
nanderstabel: EDIT: resolved via huddle

I would suggest to move this logic to the `actions.rs` file by introducing a new `run` method to either `ActionType` or `Action` (last would be my preferred choice since `Action::run()` sounds more correct to me than `ActionType::run()`.

example:
```rust
    action.run(app_handle, &app_state).await?;
```

For `Action` it would could look something like this:
```rust
impl Action {
    pub async fn run(
        &self,
        app_handle: tauri::AppHandle,
        app_state: &tauri::State<'_, AppState>,
    ) -> anyhow::Result<(), String> {
        match self.r#type {
            ActionType::GetState => {
                let transfer_state: TransferState = load_state(app_handle).await.unwrap_or_default();

                // TODO: find a better way to populate all fields with values from json file
                *app_state.status.lock().unwrap() = transfer_state.status;
                *app_state.active_profile.lock().unwrap() = transfer_state.active_profile;
                *app_state.locale.lock().unwrap() = transfer_state.locale;
            }
            ActionType::Reset => {
                reset_state(app_state.inner(), self.clone()).unwrap();
                delete_state(app_handle).await.unwrap();
            }
            ActionType::CreateNew => {
                create_did_key(app_state.inner(), self.clone()).unwrap();
                save_state(app_handle, TransferState::from(app_state.inner()))
                    .await
                    .unwrap();
            }
            ActionType::SetLocale => {
                set_locale(app_state.inner(), self.clone()).unwrap();
                save_state(app_handle, TransferState::from(app_state.inner()))
                    .await
                    .unwrap();
            }
        };
        Ok(())
    }
}
```

src-tauri/src/state/persistence.rs
nanderstabel: Actually, this seems to do the trick:
```rust
    let state_file_path = app_handle.path().data_dir()?.join("com.tauri.dev").join("state.json");
```

nanderstabel:
```suggestion
pub async fn load_state(app_handle: tauri::AppHandle) -> anyhow::Result<TransferState> {
    let state_file_path = app_handle.path().data_dir()?.join("com.tauri.dev").join("state.json");
```

src-tauri/src/state/persistence.rs
```diff
@@ -0,0 +1,50 @@
...
+    let bytes = read(&state_file_path).await?;
+    let content = String::from_utf8(bytes)?;
+    let transfer_state: TransferState = serde_json::from_str(&content).unwrap();
+    info!("state loaded from app_data_dir: {}", state_file_path.display());
+    Ok(transfer_state)
+}
+
+/// Persists a [TransferState] to the app's data directory.
+pub async fn save_state(_app_handle: tauri::AppHandle, transfer_state: TransferState) -> anyhow::Result<()> {
+    //TODO: use tauri::path::BaseDirectory::AppData;
```
nanderstabel:
```suggestion
pub async fn save_state(app_handle: tauri::AppHandle, transfer_state: TransferState) -> anyhow::Result<()> {
    let state_file_path = app_handle.path().data_dir()?.join("com.tauri.dev").join("state.json");
```

src-tauri/src/state/persistence.rs
```diff
@@ -0,0 +1,50 @@
...
+    file.write_all(serde_json::to_string(&transfer_state).unwrap().as_bytes())
+        .await?;
+    info!("state saved to app_data_dir: {}", state_file_path.display());
+    Ok(())
+}
+
+// Removes the state file from the app's data directory.
+pub async fn delete_state(_app_handle: tauri::AppHandle) -> anyhow::Result<()> {
+    //TODO: use tauri::path::BaseDirectory::AppData;
+    let state_file_path = dirs_next::data_dir().unwrap().join("com.tauri.dev").join("state.json");
```
nanderstabel:
```suggestion
pub async fn delete_state(app_handle: tauri::AppHandle) -> anyhow::Result<()> {
    let state_file_path = app_handle.path().data_dir()?.join("com.tauri.dev").join("state.json");
```

src-tauri/src/state/state.rs
```diff
@@ -0,0 +1,72 @@
...
+use ts_rs::TS;
+
+#[derive(Default)]
+pub struct AppState {
+    pub status: Mutex<StateStatus>,
+    pub active_profile: Mutex<Option<Profile>>,
+    pub locale: Mutex<String>,
+}
+
+// TODO: design: is that StateStatus really needed?
```
nanderstabel: No, so let's remove it if there is no explicit use for it (yet).

src-tauri/src/state/reducers.rs
```diff
@@ -0,0 +1,111 @@
...
+                r#type: ActionType::SetLocale,
+                payload: Some(json!({"locale": "nl"})),
+            },
+        )
+        .unwrap();
+
+        assert_eq!(*state.status.lock().unwrap(), StateStatus::Stable);
+        assert_eq!(*state.locale.lock().unwrap(), "nl".to_string());
+    }
+
```
nanderstabel, line 58: I think it would be beneficial to add some unit-tests for the reducers. For example, imo reducers should never panic, so when they receive invalid input they should return an `Error`.

```suggestion

    #[test]
    fn test_set_locale_with_invalid_payload() {
        let state = AppState::default();

        // Assert that a `SetLocale` action without a payload returns an error.
        assert!(set_locale(
            &state,
            Action {
                r#type: ActionType::SetLocale,
                payload: None,
            },
        )
        .is_err());
        // Assert that a `SetLocale` action with an invalid payload returns an error.
        assert!(set_locale(
            &state,
            Action {
                r#type: ActionType::SetLocale,
                payload: Some(json!({"non-existing": "kv-pair"})),
            },
        )
        .is_err());
    }

```

src-tauri/src/state/reducers.rs
```diff
@@ -0,0 +1,111 @@
+use crate::state::actions::Action;
+use crate::state::state::{AppState, Profile, StateStatus};
+use tracing::{warn, info};
+
+/// Sets the locale to the given value. If the locale is not supported yet, the current locale will stay unchanged.
+pub fn set_locale(state: &AppState, action: Action) -> anyhow::Result<()> {
+    let payload = action.payload.unwrap();
```
nanderstabel: I suggest to use `Option::ok_or` to avoid panicking on invalid input an return an `Result::Err()` instead.
```suggestion
    let payload = action.payload.ok_or(anyhow::anyhow!("unable to read payload"))?;
```

src-tauri/src/state/reducers.rs
```diff
@@ -0,0 +1,111 @@
...
+    let locale = payload["locale"].as_str().expect("unable to read locale from json payload");
```
nanderstabel: Using `expect()` will also result in a panic when there's invalid input. You can use `ok_or()` here as well.
```suggestion
    let locale = payload["locale"]
        .as_str()
        .ok_or(anyhow::anyhow!("unable to read locale from json payload"))?;
```

src-tauri/src/state/reducers.rs
```diff
@@ -0,0 +1,111 @@
...
+        *state.locale.lock().unwrap() = locale.to_string();
+        info!("locale set to: `{}`", locale);
+    }
+    Ok(())
+}
+
+/// Creates a new profile with a new DID (using the did:key method) and sets it as the active profile.
+pub fn create_did_key(state: &AppState, action: Action) -> anyhow::Result<()> {
+    let payload = action.payload.unwrap();
+    let display_name = payload["display_name"].as_str().expect("unable to read display_name from json payload");
```
nanderstabel: idem here, I would suggest `ok_or()`.

src-tauri/src/command.rs
nanderstabel: I would say we need to avoid the use of `unwrap()` and `expect()` as much as possible. 

Especially in the design pattern we want to use, if any invalid input is encountered in any of the reducers, then it should simply propagate a `Result::Error(message)` all the way back here. Since on invalid input we choose to emit the current state, we can just ignore the the `Result` and continue emitting the `app_state`.

For example:
```rust
        ActionType::SetLocale => {
            // Inner block will only be executed if `set_locale()` returns a `Result:Ok`.
            if set_locale(app_state.inner(), Action { r#type, payload }).is_ok() {
                save_state(app_handle, TransferState::from(app_state.inner()))
                    .await
                    // any `Result` will just simply be ignored.
                    .ok();
            }
        }
```

src-tauri/Cargo.toml
```diff
@@ -1,29 +1,36 @@
...
 serde_json = "1.0"
 tracing = "0.1.36"
 tracing-subscriber = "0.3.15"
 tokio = { version = "1.26.0", features = ["macros"] }
 ts-rs = "6.2"
+anyhow = "1.0"
+dirs-next = "2.0.0" # TODO: resolves the app data directory, use tauri::path::BaseDirectory::AppData instead
+
+[dev-dependencies]
+mockall = "0.11"
```
nanderstabel: This this dependency is not being used (yet) so let's remove it for now.

src-tauri/src/command.rs
```diff
@@ -1,39 +1,61 @@
...
+
+    let updated_state = TransferState::from(app_state.inner());
+    emit_event(window, updated_state.clone()).unwrap();
```
nanderstabel: `updated_state` does not need to be cloned, an we can ignore the `emit_event` `Result`.
```suggestion
    emit_event(window, updated_state).ok();
```

src-tauri/src/state/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod actions;
+pub mod persistence;
+pub mod reducers;
+pub mod state;
```
nanderstabel: Minor remark: 

This `state` module has the same name as it's parent module which is not the conventional way of using the rust module system: https://rust-lang.github.io/rust-clippy/master/index.html#module_inception.

No big deal at all, but you might want to rename the parent module, or move all your code in `state.rs` into this `mod.rs` file.

nanderstabel (changes requested):
I think this is a great base for our wallet repo to start of with! The state management in the backend "feels" very clean and structured.

I've added a suggestion for using tauri's path resolver, so that we don't need to use the `dirs_next` crate: `app_handle.path().data_dir()?.join("state.json");`

Most of my comments are related to the use of `unwrap()` in non-test code. We need to come up with a structured way of how we want to deal with certain types of errors, and how we want to handle them. 

I don't think there's one single way of doing that but let's have a discussion on that :) 

I left some ideas in the comments about but in general this is how I look at it:
- Production code should never panic, using Option and Result is always preferred.
- If it for whatever reason is not possible to use Option, or Result, then the use of `expect()` is preferred over `unwrap()`
- If more detailed messaging of specific code is needed, then use tracing, messages using `info`, `warn`, `debug`, `trace`, etc which can help out with debugging in a more fine-grained fashion.

Nonetheless, great work!

src-tauri/src/state/actions.rs
daniel-mader: Yes, we should consider refactoring this once we have some more Actions.

nanderstabel (commented):
Looks good to me :smile:

nanderstabel (approved):

Issue #7, opened 2023-04-19
Implement `did:key` method


daniel-mader:
Generate a new DID (using the `did:key` method) on first start. Otherwise load if exists.

PR #9, opened 2023-04-25
feat: add did:key method, add pre-defined dev profile

daniel-mader:
# Description of change
Creates a new keypair and a DID document for the `did:key` method. If keypair exists, use it. `dev mode` is enhanced to load an existing keypair (along with its DID) for testing purposes.

* #7 

* n/a


src-tauri/src/lib.rs
```diff
@@ -11,6 +16,26 @@ pub fn run() {
...
+    pub static ref UNSAFE_STORAGE: Mutex<std::path::PathBuf> = Mutex::new(std::path::PathBuf::new());
+}
+
+/// Initialize the storage file paths.
+fn initialize_storage(app_handle: tauri::AppHandle) -> anyhow::Result<()> {
+    *STATE_FILE.lock().unwrap() = app_handle.path().data_dir()?.join("com.tauri.dev").join("state.json");
+    info!("STATE_FILE: {}", STATE_FILE.lock().unwrap().display());
+
+    *UNSAFE_STORAGE.lock().unwrap() = app_handle.path().data_dir()?.join("com.tauri.dev").join("unsafe.bin");
+    info!("UNSAFE_STORAGE: {}", UNSAFE_STORAGE.lock().unwrap().display());
```
nanderstabel: This paths will still result in an error on `android`, My guess is because it is not possible to create a file from a `.../<folder>/<file>` PathBuf when the <folder> part does not exist in the FS yet. Removing the `.join("com.tauri.dev"` will result in this on the different platforms:
```
macos: "~/Library/Application Support/unsafe.bin"
linux: "~/.local/share/unsafe.bin"
android: "/data/user/0/com.impierce.identity_wallet/unsafe.bin"
```

On android there's already a nice project specific folder generated so I suggest the conditional path assignments below where the extra `.join()` is omitted for android. Also,  `"com.tauri.dev"` is just a generic tauri example project name, so I would suggest to use the android folder name instead: `"com.impierce.identity_wallet"`.

Btw I assume that folder name is based on the `tauri.conf.json` file so it would probably even better if we can find out a way to access the variables in that file instead of hardcoding the folder name.

```suggestion
    if cfg!(target_os = "android") {
        *STATE_FILE.lock().unwrap() = app_handle.path().data_dir()?.join("state.json");
        *UNSAFE_STORAGE.lock().unwrap() = app_handle.path().data_dir()?.join("unsafe.bin");
    } else {
        *STATE_FILE.lock().unwrap() = app_handle.path().data_dir()?.join("com.impierce.identity_wallet").join("state.json");
        *UNSAFE_STORAGE.lock().unwrap() = app_handle.path().data_dir()?.join("com.impierce.identity_wallet").join("unsafe.bin");
    }
    info!("STATE_FILE: {}", STATE_FILE.lock().unwrap().display());
    info!("UNSAFE_STORAGE: {}", UNSAFE_STORAGE.lock().unwrap().display());
```

nanderstabel: We can consider using these config parameters.
```rust
// "identity-wallet"
let product_name = app_handle
        .config()
        .package
        .product_name
// "com.impierce.dev"
let  identifier = project_dir = app_handle
        .config()
        .tauri
        .bundle
        .identifier;
```

nanderstabel (changes requested):
Great work with this useful addition of both the `initial_storage` and `load dev profile` :clap:

nanderstabel (dismissed):

nanderstabel (approved):

PR #74, opened 2023-11-09
feat: init error handling

nanderstabel:
# Description of change
Adds initial error handling by introducing `AppError` using the thiserror crate.

This PR also applies some slight refactoring of `handle_action_inner` by removing all occurrences of `save_state` and moving it to `handle_action`, which makes the code a bit cleaner. 

Instead of all the different occurrences of `save_state` we can just 'save' an unaltered copy of the app_state like so:
```rust
    // Keep a copy of the state before it is altered, so that we can revert to it if the state update fails.
    let unaltered_state = app_state.clone();
```
If `handle_action_inner` returns an `Ok(())`, then we know the state was successfully updated and thus we know we want to save the updated state (and emit it to the frontend).

Else, if `handle_action_inner` returns an `AppError`, we want to discard all preliminary updates to the `AppState` that may have occurred by simply continuing with using the `unaltered_state`. However, we still want to add the Error message to the `debug_messages` before saving the 'unaltered_state` and emitting to the frontend:

```rust
        Err(error) => {
            // If the state update fails, we revert to the unaltered state and add the error to the debug messages.
            {
                let debug_messages = &mut unaltered_state.debug_messages.lock().unwrap();
                while debug_messages.len() > 1000 {
                    debug_messages.remove(0);
                }
                debug_messages.push_back(format!(
                    "{} {:?}",
                    chrono::Utc::now().format("[%Y-%m-%d][%H:%M:%S]"),
                    error
                ));
            }
            warn!("state update failed: {}", error);

            // Save and emit the unaltered state including the error message.
            save_state(&unaltered_state).await.ok();
            emit_event(window.clone(), &unaltered_state).ok();
        }
```

n/a

Two tests added:
- `test_qr_code_scanned_invalid_qr_code_error`
- `test_get_state_unlock_storage_invalid_password`

~~- [ ] I have made corresponding changes to the documentation~~

### DIFF
--- a/src-tauri/tests/tests/get_state.rs
+++ b/src-tauri/tests/tests/get_state.rs
@@ -10,7 +10,7 @@ async fn test_get_state_create_new() {
     setup_state_file();
     setup_stronghold();
 
-    // Deserializing the Transferstates and Actions from the accompanying json files.
+    // Deserializing the AppStates and Actions from the accompanying json files.
     let state1 = json_example::<AppState>("tests/fixtures/states/no_profile_redirect_welcome.json");
     let state2 = json_example::<AppState>("tests/fixtures/states/redirect_me.json");
     let action1 = json_example::<Action>("tests/fixtures/actions/get_state.json");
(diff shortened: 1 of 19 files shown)


Oran-Dan (commented):
I don't know the implications for the front-end, but from my point of view it looks like a great improvement. Very explicit and intuitive error handling. And the fn handle_action_inner looks alot better too now.

src-tauri/tests/tests/qr_code_scanned.rs
```diff
@@ -144,3 +144,43 @@ async fn test_qr_code_scanned_handle_oid4vp_authorization_request() {
...
+    assert_state_update(
+        // Initial state.
+        AppState {
+            active_profile: Mutex::new(active_profile.clone()),
+            managers,
+            ..AppState::default()
+        },
+        // A QR code is scanned containing an invalid payload.
+        vec![action],
+        // The state is updated with a new user prompt containing the credential offer.
```
daniel-mader: Update comment

src-tauri/tests/tests/get_state.rs
```diff
@@ -70,3 +70,41 @@ async fn test_get_state_unlock_storage() {
...
+        },
+        vec![
+            // Get the initial state.
+            action1, // Unlock the storage.
+            action2,
+        ],
+        vec![
+            // The storage is locked, so the user is prompted to unlock it.
+            Some(state1),
+            // The storage is unlocked, so the user is redirected to the profile page.
```
daniel-mader: Update comment

daniel-mader (approved):
Awesome! This greatly enhances app insights! 🕵️

PR #83, opened 2023-12-06
refactor: removing all the mutexes from inside the `AppState` struct

Norlock:
# Description of change
The mutex is now moved up one level. so nowhere in the appstate there needs to be locking of mutexes anymore.

The unaltered_state is temporarily uncommented and just used the existing struct. But I doubt this have ever worked because the whole code was behind mutexes so I guess it would only clone the pointer the the referenced data. But we maybe need to tackle this in another PR or this was implemented somewhere and I missed in which case I will update it. @nanderstabel 

Ran through the demo, until the main screen.

### DIFF
--- a/src-tauri/src/command.rs
+++ b/src-tauri/src/command.rs
@@ -12,45 +12,46 @@ use crate::state::reducers::{
     update_profile_settings,
 };
 use crate::state::user_prompt::CurrentUserPrompt;
-use crate::state::AppState;
-use log::{info, warn};
+use crate::state::{AppState, AppStateContainer};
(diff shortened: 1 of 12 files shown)


nanderstabel (approved):
Great simplification of the `AppState`, the code looks much cleaner now! Also thank you for pointing out that the 'unaltered_state' does not have any effect.

It requires a bit of planning of how we can/want to make sure the `AppState` won't be updated when an error occurs (apart from updating `debug_messages`), so I agree that this would better fit in a new PR.

Great work!

Issue #60, opened 2023-10-03
Add Search to Credentials & Connections 


JelleMillenaar:
### Description

Add a Search functionalities to the Credentials and Connections pages. The search should feel responsive but doesn't search immediately after each key stroke. Searching should work based on Issuer, Credential Name or Credential Content. 

### Motivation

Improve scalability of the app by finding credentials and connections more easily if we assume large amount of usage. 

### Resources

Example from History Designs: 
![image](https://github.com/impierce/identity-wallet/assets/29886335/3e3c9c73-edc5-4f9d-a2bd-1cea68edb7e5)

### To-do List

- [x] Create Search Logic & Components
- [x] Add functionalities to "Me"
- [x] Add functionalities to "Activity" - Connections

Oran-Dan:
To-Do List - Oran
- [x] Create Search Logic & Components
- [x] Create Sort Logic & Components

SoulaAndrikopoulos:
To-Do List - Soula
- [x] Create Search Logic & Components/
- Iterate on search component's designs
- Implement and build components
 - [x] Create Sort Logic & Components
 - Iterate on sort component's designs
- Implement and build components
 - [x] Add functionalities to "Me"
 - [x] Add functionalities to "Activity" - Connections

SoulaAndrikopoulos:
Add sort component, currently without the grid/list view, and add persistent sorted preference option [see issue 72](https://github.com/impierce/identity-wallet/issues/72).
![★ V2 Dashboard](https://github.com/impierce/identity-wallet/assets/43676124/4d82a468-674c-4226-8da5-488e6338887b)

Oran-Dan:
As discussed in weekly sync, Search Page will be one and the same for connections, credentials and possibly history in the future. To avoid one big pile of search results we will add chips/filters to the search page functionality as specified in issue
Add Chips to Search Functionality #85

PR #86, opened 2023-12-08
feat: add search for credentials

SoulaAndrikopoulos:
# Description of change
N/A

Add Search & Sorting to Credentials and Connections #60 

N/A


src-tauri/src/state/reducers/user_data_query.rs
daniel-mader, line 1: I think we can extract a generic "search for this string in this array of strings" function to reduce some repetetive code. This can also be cleanly tested and wouldn't require that many json files. Generally, I think deserializing `AppState` from a json for each sorting & searching test is a bit overkill, we could easily use a default here and keep it in pure Rust as well. None of this is critical and we can make it part of a refactoring (for example once we introduce feature states). 🙂

daniel-mader (dismissed):
Great work from both of you! 🥳 One remark about the Rust code: I think for the tests we can extract a generic "search for this string in this array of strings" function to reduce some repetitive code. This can also be cleanly tested and wouldn't require that many json files. Generally, I think deserializing `AppState` from a json for each sorting & searching test is a bit overkill, we could also use a default here and keep it in pure Rust as well. However, none of this is critical and we can make it part of a refactoring (for example once we introduce feature states). 🙂

nanderstabel (dismissed):
Great Work!

daniel-mader (approved):

PR #92, opened 2024-01-12
refactor: change `enum Action` to `Arc<dyn ActionTrait>`

### LINKED ISSUE #97: Refactor `enum Action` into `trait ActionTrait`
### Description

The goal of this issue is to restructure Actions and Reducers in such a manner that there is a clear format to adhere to. This will make it much clearer and easier for Impierce but also for third parties to add and navigate through the Actions and Reducers.

The idea is the have all Actions as structs with traits so that the Reducers will be coupled to the Actions as implementations.

The Reducers will remain in their separate file structure, as they currently are.

We will also use the `typetag` library to compliment these structs. This will make it possible to deserialize Actions from front-end payloads only by name of the struct. This will keep communication between front-end and back-end mostly as it was.

The files will be restructured as well.
One folder for actions and one file per action.
The mod file will contain the format (trait) specifications as well, supplemented with guidelines for third party usage.
The result will be a clearer structure of our own code but most certainly also a much clearer, cleaner and safer pattern to follow for third parties to add their code.

### Motivation

This will make sure we have a cleaner solution for both `Action`s as well as `Reducer`s that's closer to the Redux pattern. 

Implementing Actions with the help of the `typetag` crate will also allow for adding custom Actions 'from the outside', thus allows for extensibility.

### Resources

n/a

### To-do List

[ ] Define traits & macro rules in mod file
[ ] Transpose existing code to new format
[ ] Create file structure and add comments
[ ] Add guidelines to the mod file
[ ] Test new code

nanderstabel:
# Description of change
This change introduces the `trait ActionTrait`:
```rust
/// A trait that all Actions must implement.
#[typetag::serde(tag = "type", content = "payload")]
pub trait ActionTrait: Send + std::fmt::Debug + DowncastSync {
    /// Returns the reducers that should be called when this action is dispatched.
    fn reducers<'a>(&self) -> Vec<Reducer<'a>>;
}
```
This trait is implemented on all types that represent `Action`s. Currently our Actions are represented inside a `enum Action`, which means that the set of Actions is bounded (bounded by the size of the enum). The new `ActionTrait` trait (in combination with the `typetag` crate) allows for an unbounded amount of Actions.

This allows for the `handle_action` command handler to still be able to deserialize actions:
```rust
pub async fn handle_action<R: tauri::Runtime>(
    action: Action,                                     // <-- `Arc<dyn ActionTrait>`
    _app_handle: tauri::AppHandle<R>,
    container: tauri::State<'_, AppStateContainer>,
    window: tauri::Window<R>,
) -> Result<(), String>
```

The `ActionTrait` trait can be implemented like this:
```rust 
#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone)]
#[ts(export, export_to = "bindings/actions/UnlockStorage.ts")]
pub struct UnlockStorage {
    pub password: String,
}

#[typetag::serde(name = "[Storage] Unlock")]
        vec![reducer!(unlock_storage)]
    }
}
```

Where the `fn reducers` method returns a vector of function pointers pointing to specific reducer functions.

The action reducers 'listen' to incoming actions by trying to downcast the generic `Arc< dyn ActionTrait>` to an action implementation of that trait, for example:
```rust
pub async fn unlock_storage(state: AppState, action: Action) -> Result<AppState, AppError> {
    if let Some(password) = listen::<UnlockStorage>(action).map(|payload| payload.password) {
        ...
       return Ok(updated_state);
   }
   Ok(state)
}
```
If the incoming action cannot be downcasted, than the current state is returned without being updated.

So the other notable change to the reducer functions is that their prototype changed from:
```rust
pub async fn reducer_function(state: &mut AppState, action: Action) -> Result<(), AppError>
```

to:
```rust
pub async fn reducer_function(state: AppState, action: Action) -> Result<AppState, AppError>
```
This means that a reducer always takes the current state, and only returns an updated state when the entire reducer completes successfully. Otherwise, the same current unchanged state is returned. This solves the issue of the appstate not correctly being 'reset' first mentioned in #83.

Closes #97 

Existing tests still pass, successful manual testing on physical testing with the NGDIL demo.


nanderstabel:
Closing this PR since it's merged into this PR now: https://github.com/impierce/identity-wallet/pull/102

Edit: opened again

src-tauri/identity_wallet/src/state/reducers/mod.rs
```diff
@@ -4,225 +4,282 @@ pub mod dev_mode;
...
-        state
-            .active_profile
-            .as_mut()
-            .unwrap()
-            .picture
-            .replace(picture.to_string());
-        debug!("updated picture: {}", picture);
-    });
+pub async fn update_profile_settings(state: AppState, action: Action) -> Result<AppState, AppError> {
+    if let Some(UpdateProfileSettings { theme, name, picture }) = listen::<UpdateProfileSettings>(action) {
```
daniel-mader, line 228: Since all fields in the action are optional (theme, name, picture) and the frontend only provides the field that was changed, you now always reset the profile to default for all values except the actual changed field. Existing fields should keep the same value as in `.replace()`. We should also add a test for it.

src-tauri/tests/fixtures/actions/authenticate_connect_accept.json
```diff
@@ -1,3 +1,4 @@
 {
-    "type": "[Authenticate] Connection accepted"
+    "type": "[Authenticate] Connection accepted",
+    "payload": null
```
daniel-mader: The tests also run if I remove the `payload` completely. Can we remove `"payload": null` in all test fixtures?

src-tauri/identity_wallet/src/command.rs
```diff
@@ -149,20 +37,39 @@ pub async fn handle_action<R: tauri::Runtime>(
...
+            user_data_query: guard.user_data_query.clone(),
+            dev_mode_enabled: guard.dev_mode_enabled,
+        },
+        action,
+    )
+    .await
+    {
+        // If the state update succeeds, we replace the old state with the new one.
+        Ok(app_state) => {
+            println!("APP STATE: {app_state:?}\n\n");
```
daniel-mader: Can we modify the print here? We're not modifying console outputs anywhere else in the app through stuff like `\n\n`. Can we use tracing instead like `debug!("{app_state:?}")`?

src-tauri/identity_wallet/src/state/actions/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Serialize, Deserialize, TS)]
+    #[serde(tag = "type")]
+    #[ts(export, export_to = "bindings/actions/Action.ts")]
+    pub enum Action {
+        #[serde(rename = "[App] Get state")]
+        GetState {
+            #[ts(optional)]
+            payload: Option<GetState>,
```
daniel-mader: Since `GetState` never takes a `payload`, we can also remove it completely, right? I still satisfies the `ActionTrait` and would allow to offer all three types of payloads that can exist: `payload: Foobar` (mandatory), `payload: Option<Foobar>` (optional), no payload at all.

src-tauri/identity_wallet/src/state/actions/get_state.rs
```diff
@@ -0,0 +1,17 @@
+use crate::reducer;
+use crate::state::actions::ActionTrait;
+use crate::state::actions::Reducer;
+use crate::state::reducers::get_state;
+
+use ts_rs::TS;
+
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/GetState.ts")]
```
daniel-mader: No need to export bindings for an empty struct (exported to `export type GetState = null;` which doesn't make sense).

src-tauri/identity_wallet/src/state/actions/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// Downcast an Action to a specific type, if possible.
+pub fn listen<T: ActionTrait + Clone>(action: Action) -> Option<T> {
+    action.downcast_arc::<T>().ok().map(|action| (*action).clone())
+}
+
+// TODO: remove this once we have a better way to export the TS types.
+mod bindings {
+    use super::*;
+    use crate::state::actions::get_state::GetState;
```
daniel-mader: Why this import?

src-tauri/identity_wallet/src/state/actions/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+        CancelUserFlow {
+            #[ts(optional)]
+            payload: Option<CancelUserFlow>,
+        },
+        #[serde(rename = "[DEV] Set dev mode")]
+        SetDevMode { payload: SetDevMode },
+        #[serde(rename = "[DEV] Load profile")]
+        LoadDevProfile {
+            #[ts(optional)]
+            payload: Option<LoadDevProfile>,
```
daniel-mader: Payload can be removed entirely here as well

src-tauri/identity_wallet/src/state/actions/cancel_user_journey.rs
```diff
@@ -0,0 +1,17 @@
+use crate::reducer;
+use crate::state::actions::ActionTrait;
+use crate::state::actions::Reducer;
+use crate::state::reducers::cancel_user_journey;
+
+use ts_rs::TS;
+
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/CancelUserJourney.ts")]
```
daniel-mader: All exports by `ts_rs` that produce something like `export type Foobar = null;` can be removed.

src-tauri/identity_wallet/bindings/actions/CancelUserJourney.ts
daniel-mader, line 1: These types have no real value and are only misleading. I think all bindings similar to this one (`= null;`) should not be generated.

src/routes/prompt/accept-connection/+page.svelte
```diff
@@ -136,14 +136,14 @@
       label="Accept connection"
       on:click={() =>
         dispatch({
-          type: '[Authenticate] Connection accepted',
+          type: '[Authenticate] Connection accepted'
```
daniel-mader: Please run `npm run format` on the entire project (will be added to a GitHub Action soon)

daniel-mader (changes requested):

src-tauri/identity_wallet/src/state/reducers/user_data_query.rs
```diff
@@ -160,46 +175,49 @@ mod tests {
...
+        app_state = credential_query(
+            app_state,
+            Arc::new(UserDataQuery {
                 target: QueryTarget::Credentials,
                 search_term: None,
                 sort_method: Some(SortMethod::NameAZ),
-                sort_reverse: false,
-            },
+                sort_reverse: None,
+            }) as Action,
```
Norlock: Why as Action here?

src-tauri/tests/tests/credential_offer.rs
```diff
@@ -1,14 +1,19 @@
...
+use identity_wallet::{
+    oid4vci::credential_format_profiles::{
+        w3c_verifiable_credentials::jwt_vc_json::{self, JwtVcJson},
+        CredentialFormats, Parameters, WithParameters,
+    },
+    state::actions::{Action, QrCodeScanned},
+};
 use serde_json::json;
-use std::borrow::BorrowMut;
+use std::sync::Arc;
```
Norlock, line 16: please always use tokio::Arc to maintain consistency

nanderstabel: which tokio::Arc do you mean exactly?

src-tauri/identity_wallet/src/state/reducers/user_data_query.rs
nanderstabel: will remove them

daniel-mader (approved):

PR #102, opened 2024-01-22
refactor: change `enum Action` to `Arc<dyn ActionTrait>`, change file structure

### LINKED ISSUE #97: Refactor `enum Action` into `trait ActionTrait`

Norlock:
# Description of change
Directory structure has changed.

This change introduces the `trait ActionTrait`:
```rust
/// A trait that all Actions must implement.
#[typetag::serde(tag = "type", content = "payload")]
pub trait ActionTrait: Send + std::fmt::Debug + DowncastSync {
    /// Returns the reducers that should be called when this action is dispatched.
    fn reducers<'a>(&self) -> Vec<Reducer<'a>>;
}
```
This trait is implemented on all types that represent `Action`s. Currently our Actions are represented inside a `enum Action`, which means that the set of Actions is bounded (bounded by the size of the enum). The new `ActionTrait` trait (in combination with the `typetag` crate) allows for an unbounded amount of Actions.

This allows for the `handle_action` command handler to still be able to deserialize actions:
```rust
pub async fn handle_action<R: tauri::Runtime>(
    action: Action,                                     // <-- `Arc<dyn ActionTrait>`
    _app_handle: tauri::AppHandle<R>,
    container: tauri::State<'_, AppStateContainer>,
    window: tauri::Window<R>,
) -> Result<(), String>
```

The `ActionTrait` trait can be implemented like this:
```rust 
#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone)]
#[ts(export, export_to = "bindings/actions/UnlockStorage.ts")]
pub struct UnlockStorage {
    pub password: String,
}

#[typetag::serde(name = "[Storage] Unlock")]
        vec![reducer!(unlock_storage)]
    }
}
```

Where the `fn reducers` method returns a vector of function pointers pointing to specific reducer functions.

The action reducers 'listen' to incoming actions by trying to downcast the generic `Arc< dyn ActionTrait>` to an action implementation of that trait, for example:
```rust
pub async fn unlock_storage(state: AppState, action: Action) -> Result<AppState, AppError> {
    if let Some(password) = listen::<UnlockStorage>(action).map(|payload| payload.password) {
        ...
       return Ok(updated_state);
   }
   Ok(state)
}
```
If the incoming action cannot be downcasted, than the current state is returned without being updated.

So the other notable change to the reducer functions is that their prototype changed from:
```rust
pub async fn reducer_function(state: &mut AppState, action: Action) -> Result<(), AppError>
```

to:
```rust
pub async fn reducer_function(state: AppState, action: Action) -> Result<AppState, AppError>
```
This means that a reducer always takes the current state, and only returns an updated state when the entire reducer completes successfully. Otherwise, the same current unchanged state is returned. This solves the issue of the appstate not correctly being 'reset' first mentioned in #83.

Closes #97

Application been run on dev and in build, (desktop + android)
Existing tests still pass, successful manual testing on physical testing with the NGDIL demo.


unime/src-tauri/android-build/key.properties
daniel-mader, line 1: Please remove this and add it to `.gitignore`. I know it's fake values, but then add placeholders/variables instead.

unime/src-tauri/gen/android/app/proguard-rules.pro
daniel-mader, line 1: Why is this needed?

unime/src-tauri/android-build/build.gradle.kts
```diff
@@ -27,16 +19,16 @@ android {
         applicationId = "com.impierce.identity_wallet"
         minSdk = 24
         targetSdk = 33
-        versionCode = 32
-        versionName = "0.4.7"
+        versionCode = 1
```
daniel-mader: Why are we resetting the `versionCode`?

unime/src/routes/prompt/password-required/+page.svelte
```diff
@@ -78,7 +78,7 @@
           <div slot="content" class="w-full pb-[10px] pt-[20px]">
             <button
               class="h-[48px] w-full rounded-xl bg-rose-100 px-4 py-2 text-[14px]/[24px] font-medium text-rose-500"
-              on:click={() => dispatch({ type: '[App] Reset' })}>{$LL.SETTINGS.RESET_APP.CONFIRM()}</button
+              on:click={() => dispatch({ type: '[App] Reset', payload: null })}>{$LL.SETTINGS.RESET_APP.CONFIRM()}</button
```
daniel-mader: Is a payload always required or can we default to null when no payload given?

unime/src/routes/prompt/accept-connection/+page.svelte
```diff
@@ -39,7 +39,7 @@
   console.log($state.current_user_prompt);
 
   onDestroy(async () => {
-    dispatch({ type: '[User Flow] Cancel' });
+    dispatch({ type: '[User Flow] Cancel', payload: { } });
```
daniel-mader: What's the difference between `payload: { }` and `payload: null`?

unime/src/routes/(app)/me/search/+page.svelte
```diff
@@ -21,7 +21,14 @@
     <Search
       on:value={(e) => {
         searchTerm = e.detail;
-        dispatch({ type: '[User Data] Query', target: 'Credentials', search_term: e.detail });
+        dispatch({ 
+          type: '[User Data] Query',
+          payload: {
+            target: 'Credentials',
+            search_term: e.detail,
+            sort_reverse: false,
```
daniel-mader: Can `sort_reverse` have a default of `false`, so it doesn't need to be explicitly passed?

unime/src-tauri/gen/android/app/proguard-rules.pro
nanderstabel: It turned out that the reason for the android build (`cargo tauri android build`) to crash was because of this being set to `true` in the `build.gradle.kts` file:
```
        getByName("release") {
            isMinifyEnabled = true
```

Minify optimizes and shrinks the code of the apk/bundle but with the new file structure, it optimized away the `RustWebView` causing the app to crash. Adding the last three lines to the generated `proguard-rules.pro` file prevents this.

unime/src-tauri/android-build/android-init.sh
daniel-mader, line 1: I wonder if we really need such a script in the repo. It might save running a few commands once, but you shouldn't need to run `cargo tauri android init` often on your local machine anyway. I would rather refer to the official Tauri docs here.

nanderstabel: :heavy_plus_sign:

unime/src-tauri/gen/apple/ExportOptions.plist
```diff
@@ -0,0 +1,23 @@
...
+		<key>com.impierce.identity-wallet</key>
+		<string>UniMe (alpha) - Internal Test (Development)</string>
+	</dict> -->
+
+    <key>method</key>
+	<string>ad-hoc</string>
+	<key>provisioningProfiles</key>
+	<dict>
+		<key>com.impierce.identity-wallet</key>
+		<string>UniMe (alpha) - Internal Test (Distribution)</string>
```
nanderstabel: @daniel-mader , can we actually change `<key>com.impierce.identity-wallet</key>` to `<key>com.impierce.unnime</key>`  do you know?

PR #105, opened 2024-01-24
refactor: move `identity-wallet` to root directory, add `unime` directory, improve `oid4vc` imports

Norlock:
# Description of change
This will restructure the application into two directories unime and identity-wallet. Also it uses only one import for openid4vc

I have run dev build. However someone should test iOS, and I will test the android release build


unime/src/lib/credentials/CredentialDetailsDropdownMenu.svelte
nanderstabel, line 1: Since the bindings folder moved, in some of the frontend files cannot find them anymore since the path doesn't fit anymore, for example in this file:
```ts
  import type { DisplayCredential } from 'src-tauri/identity_wallet/bindings/display-credential/DisplayCredential';
```

I think the cleanest fix would be to add something like this:
```json
      "@bindings/*": ["../identity-wallet/bindings/*"]
```

to the `tsconfig.json` file. 

and replace all occurrences of `from 'src-tauri/identity_wallet/bindings/...` for `from '@bindings/...`.

unime/src-tauri/gen-static/android/build.gradle.kts
nanderstabel, line 1: What is the benefit of moving this to a new `gen-static` folder? I personally prefer it staying under version control in `unime/src-tauri/android/app/build.gradle.kts`. 

Whenever there's the need for regenerating the `gen` folder, afterwards just `git restore unime/src-tauri/gen` will get everything back into place quite conveniently imo.

nanderstabel (changes requested):
Looks good! I just got two minor comments. 

I've also tested both `android dev` as well as `android build` and both work as they supposed to :+1:

identity-wallet/Cargo.toml
```diff
@@ -6,12 +6,10 @@ rust-version = "1.75.0"
 
 [dependencies]
 tauri.workspace = true
-
-oid4vc-manager.workspace = true
-siopv2.workspace = true
-oid4vci.workspace = true
-oid4vc-core.workspace = true
-oid4vp.workspace = true
+oid4vc.workspace = true
```
daniel-mader, line 9: Should `identity-wallet` generally have workspace dependencies if it has a life on its own or manage its own versions?

src-tauri/gen/android/app/src/main/AndroidManifest.xml
daniel-mader, line 1: This file should not be deleted (although it's in the `gen` folder). I might actually need to double-check since the `gen` folder does not exist at all anymore in the new `unime/src-tauri` anymore, but it's now called `gen-static`?

.gitignore
daniel-mader, line 1: This should be moved to `unime` instead of copied.

unime/.gitignore
daniel-mader, line 1: We should take the time to split the `.gitignore` files into one that resides on the project root and one that only applies to `unime`.

identity-wallet/.gitignore
daniel-mader, line 1: Is that single ignore already enough for the entire Rust crate?

Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+log = "^0.4"
+serde_json = "1.0"
+did-key = "0.2.1"
+indoc = "2.0"
+rand = "0.8"
+wiremock = "0.5"
+serial_test = "2.0"
+tempfile = "3.5.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
daniel-mader: Can we remove this default comment?

unime/src-tauri/gen-static/android/proguard-rules.pro
daniel-mader, line 1: How did this end up here? I don't think we need to track that file at all.

identity-wallet/src/lib.rs
```diff
@@ -6,11 +6,7 @@ pub mod utils;
 pub mod verifiable_credential_record;
 
 // Re-exports
-pub use oid4vc_core;
-pub use oid4vc_manager;
-pub use oid4vci;
-pub use oid4vp;
-pub use siopv2;
+pub use oid4vc::*;
```
daniel-mader: I think we should generally avoid wildcard imports.

unime/src-tauri/resources/journey-definition.json
daniel-mader, line 1: This is a duplicate of `identity-wallet/resources/journey-definition.json`. I am unsure which is the correct location.

unime/src-tauri/resources/ngdil.json
daniel-mader, line 1: This is a duplicate of `identity-wallet/resources/ngdil.json`. I am unsure which is the correct location.

unime/src-tauri/resources/onboarding-2.json
daniel-mader, line 1: This is a duplicate of `identity-wallet/resources/onboarding-2.json`. I am unsure which is the correct location.

unime/src-tauri/resources/onboarding.json
daniel-mader, line 1: This is a duplicate of `identity-wallet/resources/onboarding.json`. I am unsure which is the correct location.

unime/src-tauri/Cargo.toml
```diff
@@ -38,36 +18,36 @@ crate-type = ["staticlib", "cdylib", "rlib"]
...
-identity-wallet = { path = "./identity_wallet" }
+tauri = { version = "=2.0.0-alpha.21", features = [
+    "protocol-asset",
+    "rustls-tls",
+] }
+
+oid4vc.workspace = true
+log.workspace = true
+serde_json.workspace = true
+did-key.workspace = true
```
daniel-mader, line 29: `did-key` used to be under **[dev-dependencies]**. Why is it in actual dependencies now?

identity-wallet/Cargo.toml
```diff
@@ -47,10 +41,10 @@ typetag = "0.2"
...
 [dev-dependencies]
-indoc = "2.0"
-rand = "0.8"
-serial_test = "2.0"
+indoc.workspace = true
+rand.workspace = true
+wiremock.workspace = true
+serial_test.workspace = true
+tempfile.workspace = true
+
```
daniel-mader, line 49: `url` is not under **[dev-dependencies]** anymore?

daniel-mader (changes requested):

identity-wallet/Cargo.toml
daniel-mader: I suppose it makes dependency management easier while both crates still live in the same repository.

identity-wallet/Cargo.toml
daniel-mader: Was it removed completely as a dependency of `identity-wallet`?

unime/src-tauri/gen-static/android/proguard-rules.pro
nanderstabel: we do need it now. This is what prevents the crash on android builds, see: https://github.com/impierce/identity-wallet/pull/102#discussion_r1462869197

identity-wallet/Cargo.toml
Norlock: Yes keep it as much in sync as possible. It easier to maintain

src-tauri/gen/android/app/src/main/AndroidManifest.xml
Norlock: Well the dev repository had those files removed, I just added gen-static for all the files that need to be in gen. but won't risk ever changed without notice

.gitignore
Norlock: Both can have /target also both can have .vscode bloatware I will change some of other others

unime/src-tauri/gen-static/android/build.gradle.kts
Norlock: well you put everything in ignore for gen I think its better to have both. To prevent accidental recreate

identity-wallet/Cargo.toml
Norlock: I looked at it and no more references where there

identity-wallet/src/lib.rs
Norlock: Well its a reexport of a reexport but I will change it

identity-wallet/.gitignore
daniel-mader: Please also add the `# Environment variables` section to the `.gitignore` as `identity-wallet` can have env variables to set feature flags for a local environment.

daniel-mader: Nevermind, env variables are only set in `unime` folder.

daniel-mader (approved):

nanderstabel (approved):

PR #106, opened 2024-01-26
feat: prevent logging of sensitive data, fix: prevent app deadlocks

Norlock:
# Description of change
Deadlock protection and more secure logging. 

I ran the application locally


unime/src-tauri/gen-static/android/build.gradle.kts
daniel-mader, line 1: I assume you can now remove this file as it should already be present in `unime/src-tauri/gen-static/android/app` (moved to reflect generated folder structure).

unime/src-tauri/gen-static/android/proguard-rules.pro
daniel-mader, line 1: I assume you can now remove this file as it should already be present in `unime/src-tauri/gen-static/android/app` (moved to reflect generated folder structure).

unime/src/lib/safe-logging.ts
```diff
@@ -0,0 +1,28 @@
+// Because logging of svelte can end up in Tauri when stringified. This function will provide some safety
+
+const defaultBadWords = ['password', 'pass', 'token', 'auth', 'secret', 'passphrase', 'card'];
```
daniel-mader: Where is this list from? Why exactly these keywords?

unime/src/lib/safe-logging.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const sanitizeStringifyRecursively = function (object: any, replacer: any, space: any): any {
```
daniel-mader: Please add some tests for custom logic 👍

unime/src/lib/safe-logging.ts
daniel-mader, line 1: I would suggest renaming the file to something like `sensitive-logging.ts` to make it more specific that it is not about protecting the app itself (aka input sanitization), but rather respecting sensitive user data.

daniel-mader (commented):
Looking good! 🥳 Please do not forget to run `npm run format` in the `unime` folder.

nanderstabel (dismissed):
Nice work!

unime/src-tauri/gen-static/android/build.gradle.kts
Norlock: fixed

unime/src-tauri/gen-static/android/proguard-rules.pro
Norlock: fixed

unime/src/lib/safe-logging.ts
Norlock: Its from a package online, its just the keywords that are often used for sensitive data. Its not a bulletproof solution but there isn't a bullet proof solution for this as far as I can think off.

unime/src/lib/safe-logging.ts
Norlock: done

unime/src/lib/safe-logging.ts
Norlock: done

daniel-mader: I still see the file name as `secure-logging.ts` (which in my understanding has different semantics than what we actually do)

unime/src/lib/safe-logging.ts
daniel-mader: Can you add that as a comment to that list in case we ever wonder how we came up with that list?

identity-wallet/src/command.rs
```diff
@@ -90,6 +78,24 @@ pub async fn handle_action<R: tauri::Runtime>(
...
 
+pub async fn handle_action<R: tauri::Runtime>(
+    action: Action,
+    app_handle: tauri::AppHandle<R>,
+    container: tauri::State<'_, AppStateContainer>,
+    window: tauri::Window<R>,
+) -> Result<(), String> {
+    tokio::select! {
+        res = main_exec(action, app_handle, container, window) => {
+            info!("Finish invoke");
```
daniel-mader: I think this can be `debug!()`.

identity-wallet/src/command.rs
```diff
@@ -90,6 +78,24 @@ pub async fn handle_action<R: tauri::Runtime>(
...
+            res
+        }
+        _ = deadlock_safety() => {
+            error!("Operation timed out");
```
daniel-mader, line 96: Do you think we should log as a `warn!()` instead?

identity-wallet/src/command.rs
```diff
@@ -29,9 +31,13 @@ async fn reduce(state: AppState, action: Action) -> Result<AppState, AppError> {
         .await
 }
 
+async fn deadlock_safety() {
+    tokio::time::sleep(Duration::from_secs(6)).await;
```
daniel-mader: Why exactly 6 seconds? Can we extract this value to a static and give it a proper name such as `REDUCER_TIMEOUT_IN_SECONDS` or something more speaking?

unime/src/lib/safe-logging.ts
Norlock: What comment do you want eventually it just defaultBadWords we can add and remove from it?

identity-wallet/src/command.rs
Norlock: No because I like to see it being finished in the log. Not just when it started

identity-wallet/src/command.rs
Norlock: But nothing got updated so it's an error?

identity-wallet/src/command.rs
Norlock: I will make a variable to it. but 6 seconds is random like any other value would do as well

unime/src/lib/secure-logging.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  test('Check if secure logged', () => {
+    let credentialObj = {
+      name: 'John doe',
+      age: 44,
+      password: 'please_dont_log',
+    };
+
+    let sampleStr = JSON.stringify(credentialObj);
+
+    let secureStr = sanitizeStringify(credentialObj);
```
daniel-mader: I know it's "just tests", but I feel like the term _secure_ has a different meaning than _sensitive_. This PR doesn't add to the security of the application, but to keep the sensitive values out of the logs. So I think we should also use that wording accordingly.

unime/src/lib/safe-logging.ts
daniel-mader: "Its from a package online, its just the keywords that are often used for sensitive data. Its not a bulletproof solution."

identity-wallet/src/command.rs
daniel-mader: What about turning on DEBUG level in your local environment then?

identity-wallet/src/command.rs
daniel-mader: What if the remote server does not respond when we try to fetch a credential offer? Will that also end up here or is that handled somewhere else?

identity-wallet/src/command.rs
daniel-mader: Let's document this then. "This value is based on an estimated guess. Can be adjusted in case lower/higher timeouts are desired."

identity-wallet/src/command.rs
Norlock: It doesn't matter right? A warning is something that is not correct but still works like missing env var but now its using a default value for that, an error is the intended action has not been successfully executed

identity-wallet/src/command.rs
Norlock: Why this is not that much information, it just prints a line when invoked or timed out? Its always nice to be able to read the log and mentally be able to follow the flow of what happened. It just seperates the boundries of an action from the frontend. But if you really feel its spammy I will change to debug

identity-wallet/src/command.rs
daniel-mader: I see errors as something we should take care of and inspect and try to avoid. Warnings are things that did not work as expected, but it might not be critical to the app continuing to work. Maybe we would need a distinction between "internal failures" (such as unlocking secure storage) and "external failures" (such as making HTTP calls over an unreliable network). But I am okay with keeping `error!()` for now. 👍

identity-wallet/src/command.rs
daniel-mader: I just want to keep the logs clean. We should assume that invoke will work 99% of the time, so always logging "Hey, it worked as expected!" feels like a classic debug message to me that is only relevant if there's an issue you want to inspect and do need when you implement a feature.

unime/src/lib/sensitive-logging.test.ts
```diff
@@ -0,0 +1,49 @@
+import { sanitizeStringify } from './sensitive-logging';
+
+describe('secure-logging', () => {
```
daniel-mader: Please update name of test suite to `sensitive-logging` as well 👍

daniel-mader (approved):

Issue #14, opened 2023-05-12
ci: run linter and tests on each new PR


daniel-mader:
### Description

Add a GitHub Action that runs linters and all tests for both frontend (JS) and backend (Rust) and reports the results back to the open PR, so the reviewer can see them.

### Motivation

Automate code quality checks.

### To-do List

- [x] Lint & test Svelte frontend
- [ ] Lint & test Rust backend
- [ ] Adjust repo settings (report has to be successful as a merge requirement)

PR #113, opened 2024-02-13
ci: add basic QA workflow for frontend (format, lint, test)

daniel-mader:
# Description of change
Introduce a baseline QA for UniMe's frontend. Runs formatting using `prettier`, linting through `eslint` and tests using `vitest`.

#14 

Opening this PR triggered the first run 😉

### DIFF
--- a/.github/workflows/format-lint-test.yaml
+++ b/.github/workflows/format-lint-test.yaml
@@ -1,8 +1,6 @@
 name: Format, Lint, Test
 
 on:
-  push:
-    branches: [ "dev" ]
   pull_request:
     branches: [ "dev" ]
 
--- a/.github/workflows/format-lint-test.yaml
+++ b/.github/workflows/format-lint-test.yaml
@@ -1,6 +1,8 @@
 name: Format, Lint, Test
 
 on:
+  push:
+    branches: [ "dev" ]
   pull_request:
     branches: [ "dev" ]
 
--- a/.github/workflows/format-lint-test.yaml
+++ b/.github/workflows/format-lint-test.yaml
@@ -23,9 +23,9 @@ jobs:
     - name: Format
       working-directory: ./unime
       run: npm run format:check
-    - name: Lint
-      working-directory: ./unime
-      run: npm run lint
+    # - name: Lint
+    #   working-directory: ./unime
+    #   run: npm run lint
     - name: Test (frontend)
       working-directory: ./unime
       run: npx vitest run
--- a/.github/workflows/format-lint-test.yaml
+++ b/.github/workflows/format-lint-test.yaml
@@ -1,0 +1,31 @@
+name: Format, Lint, Test
+
+on:
+  push:
+    branches: [ "dev" ]
+  pull_request:
+    branches: [ "dev" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Use Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '20.x'
+    - name: Install dependencies
+      working-directory: ./unime
+      run: npm install    
+    - name: Format
+      working-directory: ./unime
+      run: npm run format:check
+    - name: Lint
+      working-directory: ./unime
+      run: npm run lint
+    - name: Test (frontend)
+      working-directory: ./unime
+      run: npx vitest run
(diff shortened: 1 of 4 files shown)


.github/workflows/format-lint-test.yaml
```diff
@@ -0,0 +1,31 @@
...
+      run: npm install    
+    - name: Format
+      working-directory: ./unime
+      run: npm run format:check
+    # - name: Lint
+    #   working-directory: ./unime
+    #   run: npm run lint
+    - name: Test (frontend)
+      working-directory: ./unime
+      run: npx vitest run
```
Norlock, line 31: npm run test?

.github/workflows/format-lint-test.yaml
```diff
@@ -0,0 +1,31 @@
+name: Format, Lint, Test
+
+on:
+  push:
+    branches: [ "dev" ]
```
nanderstabel, line 5: Pushing on dev is not possible right so do we really need this?

On the other hand, if some reason someone does actually push to dev it might be nice to make sure it is at least formatted :p

.github/workflows/format-lint-test.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Use Node.js
+      uses: actions/setup-node@v3
```
nanderstabel: There is a `actions/setup-node@v4` as well should we use the latest version or does it not matter at all?

.github/workflows/format-lint-test.yaml
daniel-mader: `npm run test` runs all tests in watch mode by default (does not terminate after successful test run). I can also add a `test:run` npm script, but I chose to execute vitest directly. It will later be replaced by `npm run coverage` which terminates _and_ adds a test coverage report.

.github/workflows/format-lint-test.yaml
daniel-mader: Yeah, I know it's somewhat redundant to run _again_ after merge to `dev`, so I can also remove it if you like. It's just another assertion that the current `dev` always works.

.github/workflows/format-lint-test.yaml
daniel-mader: I just used the defaults that the official GitHub docs suggested and did not check further. I can update though.

nanderstabel: If official Github suggests this then it's not really necessary to change that I guess.

unime/package.json
```diff
@@ -11,8 +11,9 @@
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
     "test": "vitest",
     "coverage": "vitest run --coverage",
-    "lint": "prettier . --check && eslint .",
+    "lint": "eslint .",
     "format": "prettier . --write --plugin prettier-plugin-svelte --plugin prettier-plugin-tailwindcss --plugin @ianvs/prettier-plugin-sort-imports",
+    "format:check": "prettier . --check --plugin prettier-plugin-svelte --plugin prettier-plugin-tailwindcss --plugin @ianvs/prettier-plugin-sort-imports",
```
Norlock: I don't think `--plugn prettier-plugin-svelte` is needed see: https://github.com/sveltejs/prettier-plugin-svelte not just for check but also format, since we use version 3.

.github/workflows/format-lint-test.yaml
nanderstabel: I don't think though this will trigger once a branch is merged into `dev` though. I think it is only triggered when someone directly pushes their commit from their local `dev` branch to remote. I can be mistaken though.

daniel-mader: But we should not do that anyway, right? I will just leave it out. 👍

unime/package.json
daniel-mader: Nice catch! 🥳 I had to remove all plugin declarations to make it work, otherwise when you specify at least one explicitly it will overwrite the plugins defined in `.prettierrc`.

.github/workflows/format-lint-test.yaml
daniel-mader: Reintroduced it 😄

nanderstabel (approved):

Norlock (approved):

PR #115, opened 2024-02-14
fix: various UI adjustments & fixes (badge details)

daniel-mader:
# Description of change
Solved some UI bugs (mainly on the "Badge details" page).

n/a

Manual inspection


identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -36,6 +36,12 @@ lazy_static! {
...
         })
     );
+    pub static ref EDU_BADGE: VerifiableCredentialRecord = VerifiableCredentialRecord::from(
```
Norlock, line 39: Maybe we should create a json file to include all these jwt tokens. Maybe outside the scope of this ticket but we can add a TODO comment. So it keeps this file a bit cleaner

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -251,5 +260,11 @@ async fn load_credential_images(
     let file_name = format!("{}.png", badge_id);
     let mut file = File::create(ASSETS_DIR.lock().unwrap().as_path().to_owned().join(file_name))?;
     copy(&mut image_bytes, &mut file)?;
+
+    let mut image_bytes: &[u8] = include_bytes!("../../../resources/images/edubadge-1.png");
+    let file_name = format!("{}.png", edubadge_id);
+    let mut file = File::create(ASSETS_DIR.lock().unwrap().as_path().to_owned().join(file_name))?;
```
Norlock: Why not use the file function to write (write_all)? https://doc.rust-lang.org/std/fs/struct.File.html

unime/src/lib/components/atoms/Image.svelte
```diff
@@ -55,28 +55,26 @@ Displays an image (loaded from disk) or a fallback component.
...
+      console.warn(`could not load image for id=[${id}]`);
+    }}
+    on:load={() => console.log('loading complete')}
+    data-testid="image"
+  />
+{:else}
+  <slot name="fallback">
+    <svelte:component
+      this={icons[iconFallback]}
+      class={twMerge('h-[18px] w-[18px] text-slate-800 dark:text-grey', $$props.iconClass)}
```
Norlock, line 74: Can you not add like this example below: 
```
<h1 class="header {active_class}">Hello reactive variable</h1>
```
If you agree also look at line 62

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -270,23 +270,46 @@
...
+                label="Credential Offer (multi)"
+              />
+            </div>
+            <!-- NGDIL (remote) -->
+            <div class="flex flex-col space-y-2">
+              <p class="text-[14px]/[22px] font-medium text-slate-500 dark:text-slate-300">NGDIL (remote)</p>
+
+              <Button
+                variant="secondary"
+                on:click={() =>
```
Norlock: Can we maybe put big on click function in the typescript part? To keep the html a bit more clean

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -29,13 +29,13 @@
...
-  //   let color = credential.metadata.display.color || colors.at(0);
-
   let icon: any = credential.metadata.display.icon || 'User';
   let title: string = credential.metadata.display.name || credential.data.type.at(-1);
 
   let credentialLogoUrl: string | null;
-  let issuerLogoUrl: string | null;
+
+  // TODO: add issuer id which can then be used to link to the connection
+  let issuerId: string | undefined;
```
Comment (Norlock) [inline unime/src/routes/badges/[id]/+page.svelte]: let issuerId?: string;

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -49,29 +49,18 @@
...
-  const { enrichment, ...entries } = credential.data.credentialSubject.achievement;
-  // entries['issuer'] = credential.data.issuer ?? credential.issuer_name;
-  // entries['issuanceDate'] = new Date(credential.data.issuanceDate).toLocaleString('en-US', {
-  //   dateStyle: 'long',
-  //   timeStyle: 'medium'
-  // });
+  const hiddenStandardFields: string[] = ['id', 'type', 'name', 'description', 'image'];
+  const hiddenCustomFields: string[] = ['enrichment'];
+
+  const entries = { ...credential.data.credentialSubject.achievement };
```
Comment (Norlock) [inline unime/src/routes/badges/[id]/+page.svelte:57]: I don't know if you want to modify the existing field, but if you don't maybe use structuredClone(...).

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -88,6 +77,7 @@
       {#if credentialLogoUrl}
         <img
           src={credentialLogoUrl}
+          alt=""
```
Comment (Norlock) [inline unime/src/routes/badges/[id]/+page.svelte]: Is empty, maybe just short description

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -178,19 +164,19 @@
...
           {#each Object.entries(entries) as entry}
             <div class="flex flex-col items-start px-4 py-[10px]">
               <p class="text-[13px]/[24px] font-medium text-slate-500">{entry[0]}</p>
-              <p class="w-full break-words text-[13px]/[24px] font-medium text-slate-800 dark:text-white">
-                <!-- TODO: this is a hacky way to display nested data, but also to remove enclosing quotes for regular strings -->
-                {JSON.stringify(entry[1]).slice(1, -1)}
-              </p>
+              <div class="w-full break-words text-[13px]/[24px] font-medium text-slate-800 dark:text-white">
+                <!-- TODO: this is a hacky way to display nested data -->
+                <pre class="whitespace-pre-wrap [font-family:inherit]">{JSON.stringify(entry[1], null, 2)}</pre>
```
Comment (Norlock) [inline unime/src/routes/badges/[id]/+page.svelte]: Maybe create a function with one param entry, so we can give the function a name to convey its meaning.

identity-wallet/src/state/persistence.rs
```diff
@@ -56,6 +56,18 @@ pub fn clear_assets_tmp_folder() -> Result<(), AppError> {
     Ok(())
 }
 
+/// Clears the `/assets` folder inside the system-specific data directory.
+/// This is only used when resetting the app to factory defaults.
+pub fn clear_all_assets() -> Result<(), AppError> {
+    let assets_dir = ASSETS_DIR.lock().unwrap().as_path().to_owned();
```
Norlock, line 62: Since this is a lock, don't we need to make this async and use await? Because it can technically fail.

Norlock (changes requested):

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -99,6 +106,7 @@ pub async fn load_dev_profile(_state: AppState, _action: Action) -> Result<AppSt
         DRIVERS_LICENSE_CREDENTIAL.clone().display_credential.id,
         PERSONAL_INFORMATION.clone().display_credential.id,
         OPEN_BADGE.clone().display_credential.id,
+        EDU_BADGE.clone().display_credential.id,
```
Norlock: This doesn't have to be params but can be read immediately from const.

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -178,19 +164,19 @@
         </p>
       </div>
 
-      <!-- Metadata (Table: Credential Subject) -->
+      <!-- Contents (Table: Credential Subject) -->
```
Comment (SoulaAndrikopoulos) [inline unime/src/routes/badges/[id]/+page.svelte:173]: Did we change "metadata" to "contents" for a specific reason (just curious)?

unime/src/routes/badges/[id]/+page.svelte
Comment (SoulaAndrikopoulos) [inline unime/src/routes/badges/[id]/+page.svelte]: Are we providing `alt` text here for the `img`?

unime/src-tauri/tests/fixtures/states/three_credentials_redirect_me.json
nanderstabel, line 1: This can now be renamed to `four_credentials_redirect_me.json`

identity-wallet/src/utils.rs
```diff
@@ -48,6 +46,9 @@ pub async fn download_asset(url: reqwest::Url, logo_type: LogoType, index: usize
         return Err(AppError::DownloadAborted("File size is bigger than 2MB"));
     }
 
+    // TODO: in batch offer, use format!("{}_{}.{}", logo_type, index, extension)
```
nanderstabel: Is this still a TODO?

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -49,29 +49,18 @@
...
 
-  // create entries to be shown
-  const { enrichment, ...entries } = credential.data.credentialSubject.achievement;
-  // entries['issuer'] = credential.data.issuer ?? credential.issuer_name;
-  // entries['issuanceDate'] = new Date(credential.data.issuanceDate).toLocaleString('en-US', {
-  //   dateStyle: 'long',
-  //   timeStyle: 'medium'
-  // });
+  const hiddenStandardFields: string[] = ['id', 'type', 'name', 'description', 'image'];
+  const hiddenCustomFields: string[] = ['enrichment'];
```
Comment (nanderstabel) [inline unime/src/routes/badges/[id]/+page.svelte]: Feels like something that eventually should be solved in the backend. In any case it could use a comment to add context, e.g:
```typescript
// Custom fields to hide related to NGDIL demo.
const hiddenCustomFields: string[] = ['enrichment'];
```

Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte]: I think every frontend implementation should decide for itself what information it wants to show. If a frontend wants to show the "type" somewhere in the details, why not? If we handled this in the backend, we would take "display logic" to the backend. It would need to be very clear that it is just a "suggestion" to hide these fields, so I'd rather solve that through a good comment instead. I would consider renaming from `hiddenStandardFields` to `omittedStandardFields` or `skippedStandardFields` (but both do not imply visualization whereas "hidden" does) and also add a clear comment of why those fields are omitted when constructing the view. Does that make it clearer?

unime/src/routes/badges/[id]/+page.svelte
Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte:173]: Yes, it was decided in a feedback session that "Contents" makes more sense than just the generic "Metadata".

unime/src/routes/badges/[id]/+page.svelte
Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte]: I removed it for now since we cannot use it just yet (linking to connections).

unime/src/routes/badges/[id]/+page.svelte
Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte:57]: Why? I'm not modifying anything, just unpacking through [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and assigning it to a new const.

unime/src/routes/(app)/scan/+page@.svelte
daniel-mader: Yes, absolutely. We'll eventually replace all of those external call functions with simple mocks (for UI testing) soon after the new dev profile 🐲 is merged.

unime/src/lib/components/atoms/Image.svelte
daniel-mader: Yes, that's possible but wouldn't work in all cases (CSS cascading). `twMerge()` exists for a reason. [Have a look at this example](https://github.com/dcastil/tailwind-merge/blob/v2.2.1/docs/what-is-it-for.md#thanks-but-i-prefer-to-read) in case you're interested why.

unime/src/routes/badges/[id]/+page.svelte
Comment (nanderstabel) [reply unime/src/routes/badges/[id]/+page.svelte]: I think `hiddenCustomFields` is a good name for what it is, but adding a comment would me nice :+1: 

Maybe you're right that indeed what to hide and what not is too subjective to even make suggestions about :+1:

identity-wallet/src/utils.rs
daniel-mader: Nope, removed it since it's completed. 👍

unime/src/routes/badges/[id]/+page.svelte
Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte]: We could make suggestions about it, but probably in a larger comment explaining why we feel those fields are not necessary to be displayed.

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -246,23 +233,23 @@ async fn load_credential_images(
 
     // Credentials
     let mut image_bytes: &[u8] = include_bytes!("../../../resources/images/cuddlyferris.svg");
-    let file_name = format!("{}.svg", personal_information_id);
+    let file_name = format!("{}.svg", PERSONAL_INFORMATION.clone().display_credential.id);
```
nanderstabel: PERSONAL_INFORMATION **.clone()** .display_credential.id

these don't need to be cloned, same goes for the other credentials :)

identity-wallet/src/state/reducers/dev_mode.rs
daniel-mader: I extracted a helper function and switched to `file.write_all()` 👍

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -170,6 +170,16 @@
     }
   };
 
+  const mockNgdilCredentialOffer = () => {
+    dispatch({
+      type: '[QR Code] Scanned',
+      payload: {
+        form_urlencoded:
+          'openid-credential-offer://?credential_offer_uri=https://api.ngdil-demo.tanglelabs.io/api/offers/creds/u08LmjU8lAcTwx7pLMpy0',
```
Norlock: this url will change so remove, maybe you can wait for my PR I will add steps for dev profile

daniel-mader: Actually, I would like to get this PR here merged as soon as possible since it contains important UI bug fixes. The PR #117 you mentioned is a great improvement for developer experience, but it's not crucial for most clients who just want to try out the app.

nanderstabel (dismissed):

unime/src/routes/(app)/scan/+page@.svelte
Norlock: Yes but it doesn't fix that the url is not static ;), so maybe you can wait for the other branch to be merged first and then call that function or a similar one.

daniel-mader: I agree and I removed the static URL completely 👍

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -178,19 +170,18 @@
...
         >
           {#each Object.entries(entries) as entry}
             <div class="flex flex-col items-start px-4 py-[10px]">
               <p class="text-[13px]/[24px] font-medium text-slate-500">{entry[0]}</p>
-              <p class="w-full break-words text-[13px]/[24px] font-medium text-slate-800 dark:text-white">
-                <!-- TODO: this is a hacky way to display nested data, but also to remove enclosing quotes for regular strings -->
-                {JSON.stringify(entry[1]).slice(1, -1)}
-              </p>
+              <div class="w-full break-words text-[13px]/[24px] font-medium text-slate-800 dark:text-white">
+                <pre class="whitespace-pre-wrap [font-family:inherit]">{prettyPrint(entry[1])}</pre>
```
Comment (Norlock) [inline unime/src/routes/badges/[id]/+page.svelte:183]: Last thing then I will approve this PR, maybe call it printBadgeDetails(), or something like that so we now what it prints

Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte:183]: I think a function name should reflect what it does. The name is purposely generic since it is agnostic about its input, so I would not like to make the name too specific. The comment should add the necessary context, don't you think?

```ts
// TODO: this is a simple way to display any (potentially nested) data, since we don't have a proper UI design for it yet
const prettyPrint = (object: any): string => {
  return JSON.stringify(object, null, 2);
};
```

unime/src/routes/badges/[id]/+page.svelte
Comment (daniel-mader) [reply unime/src/routes/badges/[id]/+page.svelte]: Duplicate of https://github.com/impierce/identity-wallet/pull/115#discussion_r1492391882

Norlock (approved):
LGTM

nanderstabel (approved):

PR #117, opened 2024-02-14
feat: add new dev profile with dynamic data from NGDIL demo

Norlock:
# Description of change
Besides some styling changes to the dev mode menu, it has now options to load in a dev profile:
- Dragon (automated NGDIL demo profile)
- Ferris (mocked data)

Other fixes:
- Fixes a bug where the user would be thrown out of dev mode if they resetted their profile.
- Fixes a bug where debug messages would be hidden behind the dev menu
- Remove some logs to prevent spam

NOTE: This is a big PR, also test on the branch to see if it's all working as expected. 

I ran the code with dev mode enabled and disabled.


identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,217 @@ lazy_static! {
...
+            password: PROFILE_PW.to_string(),
+        }),
+    )
+    .await
+}
+
+async fn create_new_profile(state: AppState) -> Result<AppState, AppError> {
+    let create_new = CreateNew {
+        name: "Dragon".to_string(),
+        picture: "🐲".to_string(),
```
daniel-mader: Can you replace the actual icon with `"&#x1F432"` which is the [codepoint](https://emojipedia.org/dragon-face#technical) for dragon-face?

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,217 @@ lazy_static! {
...
 }
 
-pub async fn set_dev_mode(state: AppState, action: Action) -> Result<AppState, AppError> {
-    if let Some(enabled) = listen::<SetDevMode>(action).map(|payload| payload.enabled) {
-        return Ok(AppState {
-            dev_mode_enabled: enabled,
-            current_user_prompt: None,
-            ..state
-        });
+const PROFILE_PW: &str = "sup3rSecr3t";
```
daniel-mader: I would generally try to avoid potentially ambiguous abbreviations throughout the code and rather spell it out: `PROFILE_PASSWORD` or even just `PASSWORD`.

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,217 @@ lazy_static! {
...
+        Arc::new(UnlockStorage {
+            password: PROFILE_PW.to_string(),
+        }),
+    )
+    .await
+}
+
+async fn create_new_profile(state: AppState) -> Result<AppState, AppError> {
+    let create_new = CreateNew {
+        name: "Dragon".to_string(),
```
daniel-mader: Should we give it a more fun name (such as "Drogon")? 😉

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +52,74 @@
       goto(`/prompt/${type}`);
     }
   }
+
+  const menuItemCss = 'flex flex-col content-start justify-center mr-4';
```
daniel-mader: Even if it feels like it might be a good idea to extract reused classes into a JS variable, it is actually discouraged. There's a good [article on reusing styles](https://tailwindcss.com/docs/reusing-styles) that I suggest we should follow. I always try to decide case by case among the three options: 1. just copy & paste for each element, 2. loop over elements, 3. extract component

Norlock: I don't really see where it is discouraged, the bottom example with btn-primary seems to be an almost similar pattern. I can try to make a component of it if you really want

daniel-mader: I would rather go for option 1. or 2.

daniel-mader: I think you're confusing two things here:
* The example with `btn-primary` you're referring to is a custom utility class defined through the `@apply` directive of the framework itself that allows you to group classes that are always used together in a single custom class.
* Your current solution puts the classes in a standard JavaScript string inside the component itself. We wouldn't have IntelliSense anymore, no more automatic class sorting, possibly introduce "premature abstraction".

I really do not see a problem with simply copy & pasting the styles since we are only talking about 4-5 elements (which is considered a valid option by the creators of Tailwind: "duplication like this isn't even a real problem because it's all together in one place").

If you really don't like it, we can also define an inline interface inside the component that defines all dynamic parts for a `DevModeButton` and then loop over them and just style once. Extracting to its own separate Svelte component (option 3.) for such a small button seems overkill to me in this case.

README.md
```diff
@@ -117,7 +117,6 @@ cargo tauri icon
 ```
 
 ### Troubleshooting
-
```
daniel-mader: Why was this removed? Applying the standard formatter puts the blank line back.

unime/src/routes/+layout.svelte
Norlock: Its just annoying when you make changes you have to copy again to all the other ones. And maybe you forget one and wonder why it looks weird

daniel-mader: > If you really don't like it, we can also define an inline interface inside the component that defines all dynamic parts for a DevModeButton and then loop over them and **just style once**.

Norlock: Ok looped now!

identity-wallet/src/state/reducers/dev_mode.rs
Norlock: fixed

identity-wallet/src/state/reducers/dev_mode.rs
Norlock: fixed

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +51,82 @@
       goto(`/prompt/${type}`);
     }
   }
+
+  interface DevModeButton {
+    stringIcon?: string;
+    onClick: () => void;
+  }
+
+  function createDevButtons(): DevModeButton[] {
```
daniel-mader, line 67: What's the benefit of creating a `function createDevButtons()` and then call it just once afterwards `const devButtons = createDevButtons();`?

You could just assign it directly, no? `const devButtons: DevModeButton[] = { ... }`

unime/src/routes/+layout.svelte
```diff
@@ -32,16 +27,15 @@
...
-
+  let expandDevMenu = true;
   let showDebugMessages = false;
 
-  // set color scheme
   const systemColorScheme = window.matchMedia('(prefers-color-scheme: dark)');
+
   systemColorScheme.addEventListener('change', (e) => {
     determineTheme(e.matches, $state?.active_profile?.theme);
+    systemColorScheme.removeEventListener('change', () => {});
```
daniel-mader: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once), we can improve it further by supplying `{ once }` directly to the function itself, like so: `systemColorScheme.addEventListener('change', (e) => { }, { once: true });`

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,248 @@ lazy_static! {
...
+            ProfileType::None => {}
+        }
     }
+
     Ok(state)
 }
 
-pub async fn load_dev_profile(_state: AppState, _action: Action) -> Result<AppState, AppError> {
-    info!("load dev profile");
+pub async fn toggle_dev_settings(mut state: AppState, _action: Action) -> Result<AppState, AppError> {
```
daniel-mader: The function name is now misleading, since it does not "toggle" anything anymore (which means switching between two options). All it does is set the `dev_profile` to `None` in any case. You could as well call it `disable_dev_profile()` or `clear_dev_profile()` now. Also, I think it's quite weird that there can exist a `Some(ProfileType::None)`. How does that reflect in the app exactly? Is there a `dev_profile` now or not?

Norlock: It does toggle, it turn on or off dev settings. It just sets it back to profile none after.

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,248 @@ lazy_static! {
...
+    let state = command::reduce(state, Arc::new(qr_code)).await?;
+
+    let cr_selected = CredentialOffersSelected {
+        offer_indices: vec![0],
+    };
+
+    command::reduce(state, Arc::new(cr_selected)).await
+}
+
+async fn reset_settings(state: AppState) -> Result<AppState, AppError> {
```
daniel-mader: What "settings" are reset here? It just reduces the `reset_state()` under the hood, right? Why is this extra function needed?

Norlock: I just made for every step a function to keep it consistent, its also easy to read

unime/src/routes/+layout.svelte
Norlock: Yes but this is just for readability of code

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,248 @@ lazy_static! {
...
+    } else {
+        state.dev_profile = Some(ProfileType::None);
+    }
+
+    state.current_user_prompt = None;
+
+    Ok(state)
+}
+
+pub async fn login_profile(state: AppState) -> Result<AppState, AppError> {
```
daniel-mader: I would rename this to `unlock_storage()` if that's the only thing it is doing. I consider "login" to be a multi-step process.

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,248 @@ lazy_static! {
...
+    command::reduce(
+        state,
+        Arc::new(UnlockStorage {
+            password: PASSWORD.to_string(),
```
daniel-mader, line 84: Since that `login_profile()` function is called for whenever `dev_profile != Some(ProfileType::None)`, that implies that _all_ dev profiles share the same password (which is okay since it's public anyway). We need to make sure that all dev profiles use the `const PASSWORD`.

identity-wallet/src/state/mod.rs
```diff
@@ -46,7 +46,7 @@ pub struct AppState {
     pub locale: Locale,
     pub credentials: Vec<DisplayCredential>,
     pub current_user_prompt: Option<CurrentUserPrompt>,
-    pub dev_mode_enabled: bool,
+    pub dev_profile: Option<ProfileType>,
```
daniel-mader: Through introducing this optional enum, you will now have states such as:

```json
{
  "active_profile": {
    "name": "Shenron"
  },
  "dev_profile": "Shenron"
}
```

I assume from the logic of the rest of the code, that the following state would never be produced (since it would be inconsistent):

```json
{
  "active_profile": {
    "name": "John Doe"
  },
  "dev_profile": "Ferris"
}
```

Why have a `ProfileType` that distinguishes between more types than just "yes, there is a dev profile loaded" / "no, there is no dev profile loaded"? This would make it a boolean and we could just call it `dev_profile_active` or something and automatically inject the dev profile password based on that boolean value.

daniel-mader: Another option: we could also just add another field to `active_profile` such as `dev_profile: bool` and default to `false`. This is where it would make the most sense. You would read the profile anywhere in the code and then make distinctions based on that inner flag of the profile itself. This would hurt the business requirement to "keep the dev mode out of other structures as much as possible" though.

identity-wallet/src/state/reducers/dev_mode.rs
Norlock: Yes I will make a comment in the load_dev_profile function.

identity-wallet/src/state/mod.rs
Norlock: I will make a different enum on the state to differentiate between auto login and not. So 
DevMode {
  On,
  Off,
  OnWithAutoLogin
}

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -187,7 +187,7 @@
     console.log('onMount: /scan');
     document.documentElement.querySelector('body')!!.classList.add('transparent');
     // permissionsGiven = await checkScanPrerequisites();
-    if (!$state?.dev_mode_enabled) {
+    if (!$state?.dev_profile) {
```
SoulaAndrikopoulos: Can you explain why it's not a boolean (enable/disable) anymore, and how this dev_profile works now? :)

identity-wallet/src/state/mod.rs
nanderstabel: So I guess the reason why `dev_mode_enabled` 'cannot' be a `bool`, is because we need to make the distinction between 1. a regular profile enabling `dev_mode` and a 2. Dev Profile (that per definition is in `dev_mode`).
1. When opening the app, the `get_state` reducer should set: `state.current_user_prompt = Some(CurrentUserPrompt::PasswordRequired);` so that the user is forced to enter their password to unlock stronghold, etc..
2. For Dev Profiles, there should not be a CurrentUserPrompt, instead the storage should be automatically unlocked using the `PASSWORD` const. 

To add a bit more context: `fn get_state` just loads a state, including the `active_profile`. It is totally unaware if this `active_profile` is a  regular profile in `dev_mode` or a Dev Profile in `dev_mode`. 

I think that's the core of it, the problem to be solved is: how do we distinguish between the two? (please correct me if I'm wrong).

I think @Norlock suggestion is a viable option.

Norlock: I have implemented the suggestion above

identity-wallet/src/state/reducers/dev_mode.rs
Norlock: Changed.

unime/src/routes/+layout.svelte
Norlock: Fixed

README.md
Norlock: fixed

unime/src/routes/(app)/scan/+page@.svelte
Norlock: It already changed but know it is: On / Off / OnWithAutologin. Because we need to be able to see if the user can be automatically logged in or not.

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -1,13 +1,51 @@
 use crate::reducer;
+use ts_rs::TS;
 use crate::state::actions::ActionTrait;
 use crate::state::actions::Reducer;
 use crate::state::reducers::dev_mode::load_dev_profile;
 
-#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
-pub struct LoadDevProfile;
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone, PartialEq, Eq)]
+#[ts(export, export_to = "bindings/actions/DevProfileType.ts")]
```
nanderstabel: Since this is somewhat related to Actions, but not actually an `Action` I would prefer to see this in a separate folder e.g: `bindings/dev/DevProfileType.ts`. 

Same for `ProfileSteps` :)

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -1,13 +1,51 @@
...
+    AcceptConnection,
+    /// Step 6
+    AddPresentation,
+    /// Step 7
+    ShareCredentails,
+    /// Step 8
+    AddFutureEngineer,
+    /// Step 9
+    CompleteFlow,
+}
```
nanderstabel, line 38: What is the actual use of having this enum and:
```typescript
  // Order needs to match the BE enum.
  const profileSteps: ProfileSteps[] = [
    'CreateProfile',
    'AddCredentials',
    'AcceptCredentials',
    'AddConnection',
    'AcceptConnection',
    'AddPresentation',
    'ShareCredentails',
    'AddFutureEngineer',
    'CompleteFlow',
  ];
```
in the FE? Is it really needed if `'Order needs to match the BE enum.'`, if so can you add a comment to explain?

Edit: I assume that this adds more finegrained control during testing in the FE So if that is the case indeed please add a short description of its functionality :).

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +56,115 @@
...
+      stringIcon: '📜',
+      onClick: () => (showDebugMessages = !showDebugMessages),
+    };
+
+    return [backBtn, resetBtn, ferrisBtn, dragonBtn, debugBtn];
+  }
+
+  const devButtons = createDevButtons();
+
+  // Order needs to match the BE enum.
```
nanderstabel: Please be specific about which enum you're referring to :+1:

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -1,13 +1,51 @@
...
-pub struct LoadDevProfile;
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone, PartialEq, Eq)]
+#[ts(export, export_to = "bindings/actions/DevProfileType.ts")]
+pub enum ProfileType {
+    /// Dev profile which preloads data
+    Ferris,
 
-#[typetag::serde(name = "[DEV] Load profile")]
-impl ActionTrait for LoadDevProfile {
+    /// Dev profile which recreate steps
```
nanderstabel: Very minor thing :sweat_smile:  :
"/// Dev profile which recreate**s** steps"

nanderstabel (changes requested):

JelleMillenaar:
I really love the NGDIL demo profile that you have build here. My only comment based on the last presentation is the use of the emoji's in the top bar. I'd like it if we can go back to the more professionally looking icons.

identity-wallet/src/state/actions/load_dev_profile.rs
Norlock: It first conveys what its doing, and enums are actually integers so it is a good type for partialOrd. Which I use to check which steps can be executed. I added a comment

nanderstabel (dismissed):
Nice work!

identity-wallet/src/state/reducers/dev_mode.rs
```diff
@@ -38,20 +42,104 @@ lazy_static! {
...
+}
+
+pub async fn toggle_dev_settings(mut state: AppState, _action: Action) -> Result<AppState, AppError> {
+    info!("Toggle dev settings");
+
+    if state.dev_mode != DevMode::Off {
+        state.dev_mode = DevMode::Off;
+    } else {
+        // We don't preserve if user had autologin enabled
+        // So we just put it back to default (reload profile if you want to enable autologin again)
```
daniel-mader, line 71: If I follow the logic here, it would mean that:

1. A tester wants to try out a few things in a blank app and turns on `dev_mode`
2. Loads any dev profile to get some data to play around with (it doesn't matter if it's Ferris or Dragon)
3. The tester turns off `dev_mode` again to inspect how the app would look like to a regular user
4. When the app is now closed, you're locked out because: dev profile password is required + DevMode is not OnWithAutologin

Is this correct?

identity-wallet/src/state/reducers/dev_mode.rs
daniel-mader: Can you please call it `toggle_dev_mode` to be consistent and not introduce another term: `dev_settings`?

identity-wallet/src/state/reducers/dragon_profile.rs
daniel-mader, line 1: Can you please rename it to something such as `dynamic_dev_profile.rs` that is more speaking than some arbitrary "animal" name (`dragon`)?

unime/src/routes/+layout.svelte
daniel-mader: Since we have opposing views on this, I asked AI and it says:

#### Function Return
* If you need to create multiple instances of `DevModeButton` at different places in your code, using a function ensures consistency and avoids code duplication.
* This approach is useful when the creation of `DevModeButton` instances involves complex calculations, database queries, or other dynamic processes.

#### Direct Assignment
* Assigning directly is suitable when the creation of `DevModeButton` instances is simple and doesn’t require additional logic. It is more concise and straightforward.
* If you have a fixed set of `DevModeButton` objects that you want to initialize directly, this approach is cleaner.
* It’s especially useful for static data or predefined arrays.

Since we have a predefined array with a fixed length, no creation logic and only use it in one single place, I definitely consider this to be suitable for a direct assignment.

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +59,121 @@
...
+      svelteIcon: ArrowLeft,
+      onClick: () => history.back(),
+    };
+
+    const resetBtn: DevModeButton = {
+      svelteIcon: Trash,
+      onClick: () => dispatch({ type: '[App] Reset' }),
+    };
+
+    const ferrisBtn: DevModeButton = {
```
daniel-mader: Please be consistent in naming and either spell it out: `ferrisButton` or remove it completely: `ferris` (since it is defined in a `const devButtons` and it's clear from the context). There is no need for random abbreviations such as `Btn`.

unime/src/lib/components/molecules/dialogs/BottomDrawer.svelte
```diff
@@ -118,7 +118,7 @@
   {/if}
 </div>
 
-<style>
+<style lang="css">
```
daniel-mader: Why explicitly add the default value "css"?

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +59,121 @@
...
+    'CompleteFlow',
+  ];
+
+  async function loadFerrisProfile() {
+    dispatch({ type: '[DEV] Load DEV profile', payload: { profile: 'Ferris', execute_steps: null } }).then(() => {
+      // Reload page
+      setTimeout(async () => {
+        await goto('/');
+        await goto('/me');
+      }, 500);
```
daniel-mader: Can you give a little more info why you chose a timer that executes two `goto()`'s instead of something native such as `window.location.reload()`?

unime/src/routes/+layout.svelte
```diff
@@ -106,8 +183,11 @@
 
   <!-- Dev Mode: Debug messages -->
   {#if showDebugMessages}
-    <div class="relative z-10 min-h-full w-screen bg-orange-100">
-      <p class="p-4 text-center text-xs font-semibold uppercase text-orange-800">debug messages</p>
+    <div class="relative z-10 min-h-full w-screen bg-orange-100 pt-24">
+      <p class="pb-2 pt-2 text-center text-xs font-semibold uppercase text-orange-800">debug messages</p>
+
+      <hr class="mx-8 h-0.5 border-t-0 bg-orange-800 opacity-100" />
```
daniel-mader: * Why set `opacity: 100%` here explicitly? Did you encounter an issue when not setting it?
* Why `border-top-width: 0px`? If you want the divider to be thicker, why not use a different `height`?

unime/src/routes/devProfile.ts
daniel-mader, line 1: Empty file?

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +59,121 @@
...
 
 <main class="absolute h-screen">
   <!-- Dev Mode: Navbar -->
-  {#if $state?.dev_mode_enabled}
-    {#if showDevMode}
+  {#if $state?.dev_mode !== 'Off'}
+    {#if expandDevMenu}
       <div
-        class="hide-scrollbar fixed z-20 flex w-full space-x-4 overflow-x-auto bg-gradient-to-r from-red-200 to-red-300 p-4 shadow-md"
+        class="hide-scrollbar fixed z-20 flex w-full content-center overflow-x-auto bg-gradient-to-r from-red-200 to-red-300 p-4 pt-8 shadow-md"
```
daniel-mader: Why is there a `pt-8` added? Having the dev buttons closer to the top gives the content more room. Assuming you added it to prevent overlaying the little arrow ("expand menu" button) and make the dev buttons better reachable, we could also consider increasing the size of the "expand menu" button to make it easier to interact with. I think the dev mode should take as little space as possible to not block too much of the content, but also should be convenient to interact with. That's why I aligned it more at the top and made the buttons bigger.

identity-wallet/src/error.rs
```diff
@@ -10,6 +10,8 @@ pub enum AppError {
     // Generic error (all purpose)
     #[error("Error: {0}")]
     Error(&'static str),
+    #[error("Error: {0}")]
+    DevError(String),
```
daniel-mader: Why the need for a dedicated new type? Why not use the regular generic Error?

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +59,121 @@
...
-          on:click={() => (showDebugMessages = !showDebugMessages)}><Warning class="h-6 w-6" /></button
-        >
+        {#each devButtons as btn}
+          <div class="mr-4 flex flex-col content-start justify-center">
+            <button
+              class="m-0 flex h-10 w-10 content-center rounded-full bg-red-300 p-0 text-sm
+                font-medium text-red-700 hover:outline-none hover:ring-2 hover:ring-red-700 hover:ring-opacity-60"
+              on:click={btn.onClick}
+            >
+              {#if btn.stringIcon}
```
daniel-mader: What if both `stringIcon` and `svelteIcon` are undefined? Wouldn't it make more sense to just define one `icon` that is not optional and can have both types?

daniel-mader (changes requested):

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -187,7 +187,7 @@
     console.log('onMount: /scan');
     document.documentElement.querySelector('body')!!.classList.add('transparent');
     // permissionsGiven = await checkScanPrerequisites();
-    if (!$state?.dev_mode_enabled) {
+    if ($state?.dev_mode !== 'Off') {
```
daniel-mader: This is a major logical bug. `startScan()` should be called when `dev_mode` is not enabled. You flipped the logic to "when `dev_mode` is not Off" (meaning the scanner now starts directly when `dev_mode` is enabled and does not work anymore for regular users). Didn't you notice when you tested it on your device? I assumed so since you checked the box in the Definiton of Done: _I have performed a self-review of my own code_

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -269,7 +269,7 @@
         <!-- Divider -->
         <!-- <p class="my-4 h-[1px] w-full bg-slate-200" /> -->
 
-        {#if $state?.dev_mode_enabled}
+        {#if $state?.dev_mode != 'Off'}
```
daniel-mader: Why loose equality here compared to other related changes where it's strict equality?

unime/src/routes/+layout.svelte
```diff
@@ -32,16 +31,19 @@
     dispatch({ type: '[App] Get state' });
   });
 
-  let showDevMode = false;
-
+  let expandDevMenu = true;
```
daniel-mader: Suggestion: if you feel the need to rename `show` to `expand`, I suggest being consistent in naming and call it `expandedDevMenu` (or something alike) since "expand" describes a process and would rather fit to a function and not a boolean state.

unime/src/routes/+layout.svelte
```diff
@@ -57,46 +59,121 @@
...
+      svelteIcon: ScrollText,
+      onClick: () => (showDebugMessages = !showDebugMessages),
+    };
+
+    return [backBtn, resetBtn, debugBtn, ferrisBtn, dragonBtn];
+  }
+
+  const devButtons = createDevButtons();
+
+  // Order needs to match the BE: 'ProfileSteps' enum, it needs to be the same order because every step is based upon the previous.
```
daniel-mader: Please be explicit: `BE` --> `backend`. I personally know what it means, but someone who hasn't come across that term might not. It can also not be inferred from the context here.

unime/src/routes/(app)/me/settings/app/+page.svelte
```diff
@@ -2,11 +2,11 @@
   import { goto } from '$app/navigation';
 
   import { locales } from '$lib/app/locales';
-  import { dispatch } from '$lib/dispatcher';
   import LL from '$src/i18n/i18n-svelte';
   import SettingsEntry from '$src/lib/app/settings/SettingsEntry.svelte';
   import Switch from '$src/lib/components/atoms/Switch.svelte';
   import TopNavBar from '$src/lib/components/molecules/navigation/TopNavBar.svelte';
+  import { dispatch } from '$src/lib/dispatcher';
```
daniel-mader: Did you change that on purpose from `$lib` to `$src/lib`? It doesn't really make a difference except that it's shorter and as you can see we're not consistent at all, I'm just curious if there was an intention behind it. 🙂

identity-wallet/src/command.rs
```diff
@@ -5,13 +5,13 @@ use crate::state::{AppState, AppStateContainer};
 use futures::StreamExt;
 use itertools::Itertools;
 use log::error;
-use log::{debug, info, warn};
+use log::{debug, info};
```
daniel-mader: I think this can unified with `use log::error`, right?

identity-wallet/src/error.rs
daniel-mader: Is it handled differently?

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -1,13 +1,51 @@
...
 use crate::state::actions::ActionTrait;
 use crate::state::actions::Reducer;
 use crate::state::reducers::dev_mode::load_dev_profile;
 
-#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
-pub struct LoadDevProfile;
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone, PartialEq, Eq)]
+#[ts(export, export_to = "bindings/dev/ProfileType.ts")]
+pub enum ProfileType {
+    /// Dev profile which preloads data
```
daniel-mader: `/// Dev profile with predefined static data`

identity-wallet/src/state/reducers/dev_mode.rs
Norlock: Yes, if you want to enable it again you need to load dev profile. I think this solution for now is good enough, maybe in the future we can improve this. I can do some match on username + password but I don't find that very clean either

unime/src/routes/+layout.svelte
Norlock: ... I can't really, first you tell me to do one thing which I didn't like in the first place, now you tell me to do the opposite again.

unime/src/routes/+layout.svelte
Norlock: Since this is inside a function that is called createDevButtons, I would think it doesn't really matter, but I changed it anyway.

unime/src/routes/+layout.svelte
Norlock: No because the checking of undefined in the if statement I think is more clean (vs typeof string), and also the field have one type instead of it could be either. Which is in my opinion not as clean, but I will change it for this case to just icon. So it cannot be undefined

unime/src/routes/+layout.svelte
Norlock: Opacity I removed, border-top-width looks better than the values tailwind give me 0.5 it too little h-1 is too big

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -1,13 +1,51 @@
...
+    Ferris,
 
-#[typetag::serde(name = "[DEV] Load profile")]
-impl ActionTrait for LoadDevProfile {
+    /// Dev profile which recreates steps
```
daniel-mader:
```
/// Dev profile that can dynamically interact with the NGDIL demo by executing `ProfileSteps`
```

identity-wallet/src/error.rs
Norlock: Yes that's because &'static str is already not good because it doesn't allow for dynamic messages (so maybe we should change that as well). This is done because I want to parse an error response of function to a string.

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -1,13 +1,51 @@
...
+    AddFutureEngineer,
+    /// Step 9
+    CompleteFlow,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/DevProfile.ts")]
+pub struct DevProfile {
+    pub profile: ProfileType,
+    pub execute_steps: Option<ProfileSteps>,
```
daniel-mader: How can multiple steps be executed with a single `Option<enum>`?

unime/src/routes/+layout.svelte
Norlock: Because the dev menu is still shown, in both cases the one is just collapsed. I will rename to expanded

unime/src/routes/(app)/me/settings/app/+page.svelte
Norlock: Maybe because I removed it accidentally and readded it again through editor. Changed it back now

identity-wallet/src/state/reducers/dev_mode.rs
daniel-mader: If it's now a requirement, could you also apply that change and put the `PASSWORD` in `load_ferris_profile()` in L145?

identity-wallet/src/state/actions/load_dev_profile.rs
Norlock: Because the enum represent multiple steps, it all builds on the previous steps.

identity-wallet/src/state/reducers/dragon_profile.rs
daniel-mader: I am just putting myself in the shoes of someone new to the project that finds a `dragon_profile.rs` in the reducers folder. It doesn't have any association with anything "dev" releated.

identity-wallet/src/state/actions/load_dev_profile.rs
daniel-mader: I know, but what I'm saying is that `Option<ProfileSteps>` can only hold a single step at a time, right?

Norlock: Sure I will change it, I was questioning too if this should be plural or not

identity-wallet/src/command.rs
```diff
@@ -4,14 +4,13 @@ use crate::state::persistence::save_state;
...
-use log::error;
-use log::{debug, info, warn};
+use log::{debug, info, error};
 use std::time::Duration;
 use tauri::Manager;
 
 /// This function represents the root reducer of the application. It will delegate the state update to the reducers that
 /// are listening to the action.
-async fn reduce(state: AppState, action: Action) -> Result<AppState, AppError> {
+pub async fn reduce(state: AppState, action: Action) -> Result<AppState, AppError> {
```
nanderstabel: We don't know (yet) if we want to make this fully public, probably better to make it `pub(crate)` for now :)

unime/src/routes/+layout.svelte
daniel-mader: What do you mean?

* I suggested a direct assignment since I don't see the need to define a function for such a simple case.
* You dismissed the suggestion for "readability of code" and did not update.
* So I asked AI to get a third opinion and quoted it here to give a little more context on why I think a direct assignment makes sense.

How am I "telling you to do one thing and now the opposite again"?

identity-wallet/src/error.rs
daniel-mader: You can improve `Error(&'static str)` and use that one then, right? I just wanted to know if there's a specific reason for a dedicated "DevError" since the distinction towards a "regular" Error isn't quite clear to me if it's not handled in a different way.

Norlock: Sure I will do that, but I didn't want to accidentally break something or change something which wasn't discussed.

unime/src/routes/(app)/scan/+page@.svelte
Norlock: I will fix it

unime/src/routes/+layout.svelte
daniel-mader: If you feel like the presets are not suitable, you can pass any value in brackets such as: `h-[3px]`. See the documentation on [using arbitrary values](https://tailwindcss.com/docs/height#arbitrary-values). I don't think setting the border-width to 0 pixels just because it visually fixes the problem is sustainable. The next person to work on this code will have questions.

unime/src/routes/+layout.svelte
daniel-mader: Yes, you are right. In this particular case it might not matter. I just believe that "naming things" is a very important thing when building software, so I tried to stay consistent.

identity-wallet/src/state/reducers/dev_mode.rs
daniel-mader: Alright 👍 As long as we're aware of it, it's fine. We don't have to cover all edge cases at this point.

unime/src/routes/+layout.svelte
Norlock: Sure I will do that in the future

unime/src/routes/+layout.svelte
daniel-mader: > What if both `stringIcon` and `svelteIcon` are undefined? Wouldn't it make more sense to just define one `icon` that is not optional and can have both types?

You say the checking of undefined it way more clean, but you're only doing it for `stringIcon` and in the other case you're just `<svelte:component this={btn.svelteIcon} />` without checking for undefined.

identity-wallet/src/state/actions/load_dev_profile.rs
```diff
@@ -2,12 +2,50 @@ use crate::reducer;
...
+    AcceptConnection,
+    /// Step 6
+    AddPresentation,
+    /// Step 7
+    ShareCredentails,
+    /// Step 8
+    AddFutureEngineer,
+    /// Step 9
+    CompleteFlow,
+}
```
nanderstabel, line 38: Since `ProfileSteps` is only relevant for the `Dragon` profile type, should we move this enum to `dynamic_dev_profile.rs`? I'm thinking we should do the same to `pub async fn load_dragon_profile` (move it to `dynamic_dev_profile.rs`). 

Otherwise `ProfileSteps` suggests that its functionality is broader than just the `Dragon` profile type. 

But your design does give me an idea so I'm curious what you think about it (definitely out of scope for this PR though): 

In the future if we want to support more dev profiles (that don't work with NGDIL but against some other parties), what if we make `ProfileSteps` into a trait with an `fn execute` method? All the functions you currently have in `dynamic_dev_profile.rs` e.g:
```rust

pub trait ProfileSteps {
    async fn execute(&self, state: AppState) -> Result<AppState, AppError>;
}

#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone, PartialEq, Eq, PartialOrd)]
#[ts(export, export_to = "bindings/dev/ProfileSteps.ts")]
pub enum NGDILProfileSteps {
    /// Step 1
    CreateProfile,
    /// Step 2
    AddCredentials,
    /// Step 3
    AcceptCredentials,
    /// Step 4
    AddConnection,
    /// Step 5
    AcceptConnection,
    /// Step 6
    AddPresentation,
    /// Step 7
    ShareCredentails,
    /// Step 8
    AddFutureEngineer,
    /// Step 9
    CompleteFlow,
}

impl ProfileSteps for NGDILProfileSteps {
    async fn execute(&self, state: AppState) -> Result<AppState, AppError> {
        match self {
            NGDILProfileSteps::CreateProfile => create_new_profile(state).await,
            NGDILProfileSteps::AddCredentials => add_credential(state).await,
            NGDILProfileSteps::AcceptCredentials => accept_credential(state).await,
            NGDILProfileSteps::AddConnection => add_connection(state).await,
            NGDILProfileSteps::AcceptConnection => accept_connection(state).await,
            NGDILProfileSteps::AddPresentation => add_presentation_request(state).await,
            NGDILProfileSteps::ShareCredentails => share_credentials(state).await,
            NGDILProfileSteps::AddFutureEngineer => add_future_engineer(state).await,
            NGDILProfileSteps::CompleteFlow => accept_future_engineer(state).await,
        }
    }
}

impl NGDILProfileSteps {
    pub(super) async fn create_new_profile(&self, state: AppState) -> Result<AppState, AppError> {
        let create_new = CreateNew {
            name: "Shenron".to_string(),
            picture: "&#x1F432".to_string(),
            theme: "dark".to_string(),
            password: PASSWORD.to_string(),
        };

        command::reduce(state, Arc::new(create_new)).await
    }

    pub(super) async fn add_credential(&self, state: AppState) -> Result<AppState, AppError> {
        let state = self.create_new_profile(state).await?; // <-- All steps should execute the 'previous' step first

        // URL from NGDIL demo
        let url = "https://api.demo.ngdil.com/api/starting-offer";

        let payload = serde_json::json!({
            "credentials": [
                "National ID",
                "School Course Certificate",
                "Volunteer Badge",
                "Higher Education Information Literacy Level 1",
                "Business Innovation & Interdisciplinair Samenwerken"
            ]
        });

        let response: CredentialResponse = reqwest::Client::new()
            .post(url)
            .json(&payload)
            .send()
            .await
            .map_err(|err| AppError::DevError(err.to_string()))?
            .json()
            .await
            .map_err(|err| AppError::DevError(err.to_string()))?;

        let qr_code = QrCodeScanned {
            form_urlencoded: response.uri,
        };

        command::reduce(state, Arc::new(qr_code)).await
    }

    ...all the other profile step 'methods'
```

Then this inside `pub async fn load_dev_profile`:
```rust
    if let Some(dev_profile) = listen::<DevProfile>(action) {
        // All dev profiles need to use the const PASSWORD so it can automatically unlock storage.
        match dev_profile.profile {
            ProfileType::Ferris => return load_ferris_profile().await,
            ProfileType::Dragon => return load_dragon_profile(state, dev_profile).await,
        }
    }
```

could just be
```rust
    if let Some(dev_profile) = listen::<DevProfile>(action) {
        return dev_profile.execute().await;
    }
```

Again, just an idea for a future improvement, but interested in what you think :)

nanderstabel:
When creating a new (normal) profile, then go to `Settings` > `App Settings` to enable Developer mode. Now I can select either Ferris or Shenron, both have issues:
* Ferris: Get's prompted to enter password. I can login using `sup3rSecr3t`, but when I navigate to `Scan`, it's just a blank page. Probably related to https://github.com/impierce/identity-wallet/pull/117#discussion_r1495364571?
* Shenron: Has the same issue as Ferris, but also another issue on Android: Operation timed out. (storing the credentials into Stronghold takes too long (at least on my device). In the release builds this does not seem to be a problem but still we might need to increase the `TIMEOUT_SECS` to 10 seconds for now before we have a proper solution for storing data into Stronghold).

identity-wallet/src/state/actions/load_dev_profile.rs
Norlock: I can move the enum to dynamic_dev_profile.rs, however your solution won't work because it would execute only one step. Its not a match that is done now but a lesser then comparison so it would execute also the steps before, and return that state. 

But thanks for the suggestion.

nanderstabel: Ok thanks!

It would work because of this right?:
```rust
    pub(super) async fn add_credential(&self, state: AppState) -> Result<AppState, AppError> {
        let state = self.create_new_profile(state).await?; // <-- All steps should execute the 'previous' step first
```

Norlock: No because of this:

```rust
    if ProfileSteps::AddCredentials <= steps {
        debug!("Add credentials step executed");
        state = add_credential(state).await?;
    }
```
Maybe I should explicitly give the enum a number, but it check if steps (enum 5 is larger then AddCredentials). If so it would also execute that. Enum are basically just integers ;)

nanderstabel: Using a `match` is just a personal preference but it could also be replaced with your current solution of course:
```rust

    state = reset_settings(state).await?;
    state = create_new_profile(state).await?;

    if self <= steps {
        debug!("Add credentials step executed");
        state = add_credential(state).await?;
    }

    if self <= steps {
        debug!("Accept credentials step executed");
        state = accept_credential(state).await?;
    }

    if self <= steps {
        debug!("Add connection step executed");
        state = add_connection(state).await?;
    }

    if self <= steps {
        debug!("Accept connection step executed");
        state = accept_connection(state).await?;
    }

    if self <= steps {
        debug!("Add presentation step executed");
        state = add_presentation_request(state).await?;
    }

    if self <= steps {
        debug!("Share credentials step executed");
        state = share_credentials(state).await?;
    }

    if self <= steps {
        debug!("Add future engineer step executed");
        state = add_future_engineer(state).await?;
    }

    if self <= steps {
        debug!("Accept future engineer step executed");
        state = accept_future_engineer(state).await?;
    }
```

I was more wondering if you see anything in using a trait in the future for this functionality that could also be implemented for dev profiles outside our own codebase. If so then I will open a discussion for it :)

daniel-mader (dismissed):

nanderstabel (approved):

daniel-mader (approved):

PR #122, opened 2024-02-18
refactor: credential display fields

Norlock:
# Description of change
On multiple places multiple sources are used for the display name, this PR moves that logic to the BE and are directly placed so it can always use that field. It also removes two fields which aren't used in the FE. This PR should wait until dev-mode is merged.

I ran the Dragon profile and made sure the before and after returned the same results.


identity-wallet/src/state/reducers/mod.rs
```diff
@@ -116,7 +116,7 @@ pub async fn create_identity(state: AppState, action: Action) -> Result<AppState
         let profile = Profile {
             name: name.to_string(),
             picture: Some(picture.to_string()),
-            theme: Some(theme.to_string()),
+            theme: Some(theme),
```
nanderstabel, line 119: Not sure why we're using `to_string` here :sweat_smile: Can you also remove it for `name` and `picture`?

Norlock: Ok top ik voeg deze changes toe aan de nieuwe PR goed? dan is het wat makkelijker

nanderstabel: Sure :ok_hand:

Norlock:
Will be done in: #139

PR #127, opened 2024-02-22
fix: persist issuer logo for connections, temporarily determine `connection_id` based on base64-encoded issuer name

daniel-mader:
# Description of change
* persist issuer logo when connection is accepted
* since distinction between connections is still an open topic, we agreed on using the issuer name as the identifier for now (in the near future, this needs to be changed to the DID of the issuer or the (sub)domain where the connection was made to or a combination of both)

n/a

Due to the urgency of this fix, it has only been tested locally in a desktop environment

### DIFF
--- a/identity-wallet/src/state/reducers/authorization.rs
+++ b/identity-wallet/src/state/reducers/authorization.rs
@@ -210,9 +210,11 @@ pub async fn handle_siopv2_authorization_request(state: AppState, _action: Actio
     // since we currently lack a unique identitfier to distinguish connections.
     let connection_id = base64::encode_config(&client_name, base64::URL_SAFE);
 
+    persist_asset("issuer_0", &connection_id).ok();
+
     if result.is_none() {
         connections.push(Connection {
-            id: connection_id.clone(),
+            id: connection_id,
             client_name,
             url: connection_url,
             verified: false,
@@ -220,8 +222,6 @@ pub async fn handle_siopv2_authorization_request(state: AppState, _action: Actio
             last_interacted: connection_time,
         })
     };
-
-    persist_asset("issuer_0", &connection_id).ok();
 
     drop(state_guard);
     Ok(AppState {
--- a/identity-wallet/src/state/reducers/authorization.rs
+++ b/identity-wallet/src/state/reducers/authorization.rs
@@ -73,7 +73,7 @@ pub async fn read_authorization_request(state: AppState, action: Action) -> Resu
                     "{}",
                     format!(
                         "Downloading client logo from url: {}",
-                        logo_uri.clone().unwrap().as_str()
+                        logo_uri.as_ref().unwrap().as_str()
                     )
                 );
                 if let Some(logo_uri) = logo_uri.as_ref().and_then(|s| s.parse::<reqwest::Url>().ok()) {
@@ -130,7 +130,7 @@ pub async fn read_authorization_request(state: AppState, action: Action) -> Resu
                     "{}",
                     format!(
                         "Downloading client logo from url: {}",
-                        logo_uri.clone().unwrap().as_str()
+                        logo_uri.as_ref().unwrap().as_str()
                     )
                 );
                 if let Some(logo_uri) = logo_uri.as_ref().and_then(|s| s.parse::<reqwest::Url>().ok()) {
--- a/identity-wallet/src/state/reducers/authorization.rs
+++ b/identity-wallet/src/state/reducers/authorization.rs
@@ -337,9 +337,14 @@ pub async fn handle_oid4vp_authorization_request(state: AppState, action: Action
                 connection.last_interacted = connection_time.clone();
             });
 
+        // TODO: This is a HORRIBLE solution to determine the connection_id by the non-unique "issuer name".
+        // It is a TEMPORARY solution and should only be used in DEMO environments,
+        // since we currently lack a unique identitfier to distinguish connections.
+        let connection_id = base64::encode_config(&client_name, base64::URL_SAFE);
+
         if result.is_none() {
             connections.push(Connection {
-                id: "TODO".to_string(),
+                id: connection_id,
                 client_name,
                 url: connection_url,
                 verified: false,
--- a/identity-wallet/src/state/reducers/authorization.rs
+++ b/identity-wallet/src/state/reducers/authorization.rs
@@ -3,13 +3,14 @@ use crate::{
     get_unverified_jwt_claims,
     state::{
         actions::{listen, Action, CredentialsSelected, QrCodeScanned},
+        persistence::persist_asset,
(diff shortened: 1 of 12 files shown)


nanderstabel:
I've tried to find out why, but not successful yet.. (also not sure if this issue already existed before)

I've intentionally accepted these connections at different times (minutes), but every time I accept a new connection, all existing connections show the same timestamp as the newly added connection. 

Screenshot_20240222-160953


When checking out the actual Connection view, then "First connected" and "Last connected" have their actual correct timestamps so it seems to be a frontend issue.

identity-wallet/src/state/reducers/authorization.rs
```diff
@@ -211,6 +221,8 @@ pub async fn handle_siopv2_authorization_request(state: AppState, _action: Actio
         })
     };
 
+    persist_asset("issuer_0", &connection_id).ok();
+
```
nanderstabel: Consider moving this up on top of:
```rust
if result.is_none() {
        connections.push(Connection {
            id: connection_id.clone(),
            client_name,
            url: connection_url,
            verified: false,
            first_interacted: connection_time.clone(),
            last_interacted: connection_time,
        })
    };
```

So that ` id: connection_id.clone()` does not need to be cloned.

identity-wallet/src/state/reducers/authorization.rs
```diff
@@ -200,9 +205,14 @@ pub async fn handle_siopv2_authorization_request(state: AppState, _action: Actio
             connection.last_interacted = connection_time.clone();
         });
 
+    // TODO: This is a HORRIBLE solution to determine the connection_id by the non-unique "issuer name".
+    // It is a TEMPORARY solution and should only be used in DEMO environments,
+    // since we currently lack a unique identitfier to distinguish connections.
+    let connection_id = base64::encode_config(&client_name, base64::URL_SAFE);
```
nanderstabel, line 211: Now I think about it.. it's probably better to construct the `connection_id` not just on the `client_name`, but on `connection_url` as well, since the pair will be more unique than just the `client_name`. A malicious server can just take on any `client_name` and we would 'trust' the Connection because we already might have connected with a non-malicious server before with the same `client_name`.

nanderstabel: Also, now we have a `connection_id` attached to each Connection, we can now refactor this code (using `.find(|connection| connection.id == connection_id)`):
```rust
    let result = connections
        .iter_mut()
        .find(|connection| connection.url == connection_url && connection.client_name == client_name) // <-- .find(|connection| connection.id == connection_id)
        .map(|connection| {
            connection.last_interacted = connection_time.clone();
        });
```

identity-wallet/src/state/reducers/authorization.rs
```diff
@@ -200,9 +205,14 @@ pub async fn handle_siopv2_authorization_request(state: AppState, _action: Actio
...
+
     if result.is_none() {
         connections.push(Connection {
-            id: "TODO".to_string(),
+            id: connection_id.clone(),
```
nanderstabel: Is there any implication of now using a `connection_id` here, but not in `fn handle_oid4vp_authorization_request`?:
```rust
        if result.is_none() {
            connections.push(Connection {
                id: "TODO".to_string(),
                client_name,
                url: connection_url,
                verified: false,
                first_interacted: connection_time.clone(),
                last_interacted: connection_time,
            })
        };
```

unime/src/routes/badges/[id]/+page.svelte
```diff
@@ -64,6 +64,32 @@
...
+    const name_from_credential: string | undefined = credential.data.issuer?.name;
+    const name_from_oid4vc = credential.issuer_name;
+    // We prefer the name from the credential, but fallback to the issuer name during oid4vc process
+    const issuer_name = name_from_credential ?? name_from_oid4vc;
+
+    if (issuer_name) {
+      // base64url encode
+      const connectionId = btoa(issuer_name).replace('+', '-').replace('/', '_');
+      // verify that the connection exists
+      if ($state.connections.find((c) => c.id === connectionId)) {
```
Comment (Norlock) [inline unime/src/routes/badges/[id]/+page.svelte]: Maybe you can use the some function here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

nanderstabel:
> I've tried to find out why, but not successful yet.. (also not sure if this issue already existed before)
> 
> I've intentionally accepted these connections at different times (minutes), but every time I accept a new connection, all existing connections show the same timestamp as the newly added connection.
> 
> Screenshot_20240222-160953
> 
> When checking out the actual Connection view, then "First connected" and "Last connected" have their actual correct timestamps so it seems to be a frontend issue.

So apparently this has nothing to do with new connections or updating connections. The timestamps are just updated every time the `Activity` tab is opened because of:
```typescript
        <div slot="right" class="h-full pr-2 pt-1 text-[12px]/[20px] font-medium text-slate-400">
          {new Date().toLocaleString($state.locale, {
            dateStyle: 'short',
            timeStyle: 'short',
          })}
        </div>
```
in `ConnectionsList.svelte`

identity-wallet/src/state/reducers/authorization.rs
daniel-mader: Should we do this refactoring in this PR or later as part of a general overhaul when we fix the connection ids for good?

nanderstabel: Yeah fair enough, sounds good :+1:

identity-wallet/src/state/reducers/authorization.rs
```diff
@@ -67,6 +68,19 @@ pub async fn read_authorization_request(state: AppState, action: Action) -> Resu
             info!("client_name in credential_offer: {:?}", client_name);
             info!("logo_uri in read_authorization_request: {:?}", logo_uri);
 
+            if logo_uri.is_some() {
+                debug!(
+                    "{}",
+                    format!(
+                        "Downloading client logo from url: {}",
+                        logo_uri.clone().unwrap().as_str()
```
Norlock: Maybe you can use `&logo_uri.unwrap().as_str()` so you don't have to clone

Norlock (commented):
One small thing then I think its good

identity-wallet/src/state/reducers/authorization.rs
daniel-mader: Thanks! I think once we have some Rust linting in place using clippy, such issues would be reported during builds as well 👍

Norlock (approved):

identity-wallet/src/state/reducers/authorization.rs
Norlock: Yes  clippy check is a good idea as a workflow addition

nanderstabel (approved):

PR #129, opened 2024-02-22
fix: show info when scanner `PermissionState` is not `granted`

### LINKED ISSUE #124: [Bug] Camera Permissions doesn't handle "Ask Everytime" on Android
### Description

An specific cases the "scan" button leads to a blank page without a permission prompt, nor an active camera for QR scanning. It seems to be related to the edge case of "Ask Every time" as it is labelled in Android Settings menu. When the app does not have camera permission a prompt appears with 3 options:
- "While using this app" (Android Settings: "Allow only while using the app") -> Camera always works and continues to work later
- "Only this time" (Android Settings: "Ask every time") -> Works initially, but this somehow times out. Not with an app reboot, it will continue to work. I think its related to time.
- "Don't Allow" (Android Settings: "Don't Allow") -> Blank screen. In subsequent visits it shows a "No permission" feedback properly, this should have been shown directly after refusing permission. 

The biggest issue is the inconsistency of "Only this time" which stops working after some time. I assume the permission is given in a timeframe. This should properly reprompt if that permission is timed out. 

### Hardware Specification

Operating system: Android 12
Samsung Galaxy S10e

1. Click Camera / Scan button 
2. Click "Only this time" 
3. Close the app fully and wait arbitrary long timeframe
4. Open the app on scan and be shown the blank page. If the camera opened, you didn't wait long enough.

### Expected Behaviour

Prompt is shown for permissions

### Actual Behaviour

Blank screen

### Errors

-

daniel-mader:
# Description of change
Please write a summary of your changes and why you made them.

fixed #124 

### DIFF
--- a/unime/src/routes/(app)/scan/+page@.svelte
+++ b/unime/src/routes/(app)/scan/+page@.svelte
@@ -10,6 +10,7 @@
     openAppSettings,
     requestPermissions,
     scan,
+    type PermissionState,
     type Scanned,
   } from '@tauri-apps/plugin-barcode-scanner';
   import { debug, info, warn } from '@tauri-apps/plugin-log';
@@ -23,7 +24,8 @@
   import CameraSlash from '~icons/ph/camera-slash';
 
   let scanning = false;
-  let permissions: 'granted' | 'denied' | 'prompt' | null;
+  // We temporarily introduce this type that extends `PermissionState` to handle a possible error when checking for permissions.
+  let permissions_nullable: PermissionState | null;
 
   function onMessage(scanned: Scanned) {
     debug(`scanned: ${scanned.content}`);
@@ -48,23 +50,24 @@
 
   // from example in plugin-barcode-scanner repo
   async function startScan() {
-    let permission = await checkPermissions()
-      .then((permission) => {
-        info(`Permissions to use the camera: ${permission}`);
-        return permission;
+    let permissions = await checkPermissions()
+      .then((permissions) => {
+        info(`Permissions to use the camera: ${permissions}`);
+        return permissions;
       })
       .catch((error) => {
         warn(`Error checking for permissions to use the camera: ${error}`);
-        return null; // possibly "denied"? or does that imply that the check has been successful, but was actively denied?
+        return null; // possibly return "denied"? or does that imply that the check has been successful, but was actively denied?
       });
-    permissions = permission;
 
-    if (permission === 'prompt') {
-      info('requesting permission');
-      permission = await requestPermissions();
+    if (permissions === 'prompt') {
+      info('Requesting camera permissions');
+      permissions = await requestPermissions(); // handle in more detail?
     }
 
-    if (permission === 'granted') {
+    permissions_nullable = permissions;
+
+    if (permissions === 'granted') {
       // Scanning parameters
       const formats = [Format.QRCode];
       const windowed = true;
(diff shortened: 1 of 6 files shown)


Oran-Dan (commented):
in the en/index.ts file:
"permissions" should be single = permission

in nl-NL/index.ts:
recht = toestemming

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -189,111 +117,38 @@
...
       <div
         class:invisible={scanning}
-        class="relative flex h-full flex-col items-center justify-center bg-silver p-8 dark:bg-navy"
+        class="relative flex h-full flex-col items-center justify-center space-y-4 bg-silver p-8 dark:bg-navy"
       >
-        {#if permissions === 'denied'}
-          <div class="flex flex-col items-center space-y-4">
-            <div class="rounded-lg bg-rose-100 px-8 py-4 text-rose-500">
-              {$LL.SCAN.NO_PERMISSION_1()}<br />{$LL.SCAN.NO_PERMISSION_2()}
+        {#if permissions_nullable !== 'granted'}
```
nanderstabel: Although this solution seems to work, now every time I go to the `Scan` tab, for a brief moment the "No permissions to access the camera" prompt is visible before opening the camera. So it registers a bit late that the permission is actually granted. 

I remember this being an issue before but I am not completely sure.. 

Is there some way of changing the logic that only when failing to start the scanner the prompt is shown?

daniel-mader: I believe that when the permissions are not loaded yet (but are `granted`), it would still briefly show the "no permissons given" due it `!== 'granted'` at the time of check. That's why I added a check against `null` and it seems to work now.

nanderstabel (approved):

Oran-Dan (approved):
Some variables are still called ```permissions```, plural. I'm not sure if this is on purpose.
Either way, the semantic difference is dismissable, so i approve.

unime/src/routes/(app)/scan/+page@.svelte
```diff
@@ -20,73 +21,60 @@
   import BottomNavBar from '$src/lib/components/molecules/navigation/BottomNavBar.svelte';
   import { state } from '$src/stores';
 
+  import CameraSlash from '~icons/ph/camera-slash';
+
   let scanning = false;
-  let permissions: 'granted' | 'denied' | 'prompt' | undefined = undefined;
+  // We temporarily introduce this type that extends `PermissionState` to handle a possible error when checking for permissions.
+  let permissions_nullable: PermissionState | null;
```
Norlock, line 28: maybe we can call this optionalPermissions, so we keep the terminology of rusts `Option` and typescript a bit similar

Norlock (approved):

unime/src/routes/(app)/scan/+page@.svelte
daniel-mader: Yeah, I considered this as well, but they're not really optional, but actually nullable in the current implementation. "Optional" would imply that permissions could be there or couldn't while access to the camera _always_ needs some kind of permissions. So I rather chose they are actively "nullable" until [the issue that I opened](https://github.com/tauri-apps/plugins-workspace/issues/979) is resolved and we refactor it for good to not have this inconsistent state anymore.

PR #131, opened 2024-02-22
feat: add extensions field to AppState

### LINKED ISSUE #100: Make `AppState` extensible
### Description

The goal is to make the `AppState` struct extendable.
We want to make this possible by adding an extra field called `extensions`.
This field will be a hashmap of type `<String, Box<dyn>>`.
This will allow for any size of struct to be added as an extension, indexable by extension name.

### Motivation

The motivation is to move toward an app extendable in a clear and safe way.

After this issue (and the issues regarding Actions and Reducers, and UserPrompts) third party back-end developers should have no issue extending their version of our codebase.

Further down the road lies the task to format the rest of our code to the same format, forcing us to make our codebase clearly split up in neatly isolated building blocks.

### Resources

Depends on #92 

### To-do List

- [x] Add the extension field to the `AppState`
- [x] Add impl {pub fn extensions} to Appstate_container
- [x] Create tests with simple extensions and actions to interact with the extension

Oran-Dan:
# Description of change
I added the extension field to the appstate.

Issue #100 

I added testing in the unime tests folder


identity-wallet/bindings/actions/Test.ts
Norlock, line 1: Since this interface has the name Test in it, I just want to double check if this is correct

identity-wallet/src/state/extensions/reducers.rs
```diff
@@ -0,0 +1,24 @@
+use super::actions::CustomExtensionTest;
+use super::CustomExtension;
+use crate::error::AppError;
+use crate::state::actions::{listen, Action};
+use crate::state::AppState;
+
+pub async fn test_feat_state(state: AppState, action: Action) -> Result<AppState, AppError> {
+    if let Some(test_feat_state) = listen::<CustomExtensionTest>(action) {
+        
+        let mut new_state = AppState {
```
Norlock: Maybe a bit cleaner to do `state.clone()`;

identity-wallet/src/state/extensions/reducers.rs
```diff
@@ -0,0 +1,24 @@
...
+            ..state
+        };
+
+        new_state.extensions.insert("test".to_string(), Box::new(CustomExtension{
+            name: "new".to_string(),
+            value: if test_feat_state.test_bool { "new".to_string() } else { "old".to_string() }
+        }));
+
+        return Ok(AppState {
```
Norlock: Here you can do:
```rust
return Ok(new_state);
```

Norlock (changes requested):
Overal looks good, just a small improvement on cloning

identity-wallet/bindings/actions/Test.ts
Oran-Dan: Well it was autogenerated, so i am not sure, but it very much seems in line with what is generated for other actions

identity-wallet/src/state/extensions/actions.rs
nanderstabel, line 1: Since `struct CustomExtensionTest`, `struct CustomExtension` and `fn test_feat_state` are useful for demonstrating how the `AppState` functionality can be extended, but don't serve as an actual core feature, I'd say it's better to move them over to somewhere in `unime/src-tauri/tests/`. Either directly into `tests/extensions.rs` or in some common folder :)

identity-wallet/bindings/actions/Test.ts
nanderstabel, line 1: This is autogenerated by `ts_rs`, but since `CustomExtensionTest` is for demonstration purposes and not actual functionality, we won't need bindings for this type.

identity-wallet/src/state/extensions/actions.rs
```diff
@@ -0,0 +1,19 @@
+use crate::{reducer, state::{actions::ActionTrait, extensions::reducers::test_feat_state, reducers::Reducer}};
+use ts_rs::TS;
+
+
+/// Action to test the extension field.
+#[derive(serde::Serialize, serde::Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/Test.ts")]
```
nanderstabel: This creates the bindings in `identity-wallet/bindings/actions/Test.ts`, but since we won't be needing bindings for this struct this line can be omitted :)

identity-wallet/src/state/extensions/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+pub mod reducers;
+
+// This module is soly for testing and demonstrating the extension system.
+#[derive(Debug, Serialize, Deserialize, TS, PartialEq, Default, Clone)]
+pub struct CustomExtension {
+    pub name: String,
+    pub value: String,
+}
+
+#[typetag::serde(name = "customextension")]
```
nanderstabel: Probably a bit nit-picky but I'd say best to follow conventional use of snake_case here: `custom_extension`. (If you agree this should also be changed in `unime/src-tauri/tests/fixtures/states/test_extension.json`)

identity-wallet/src/state/extensions/reducers.rs
nanderstabel: Actually also no need to clone here :smiley: 

This would already do the job: `let mut new_state = state;`.

nanderstabel (changes requested):
Looks good! Just some minor comments, and don't forget to run `cargo fmt` :)

nanderstabel (approved):

Norlock (approved):

Issue #137, opened 2024-02-23
feat: refactor AppState struct


Oran-Dan:
### Description

Introduce the FeatTrait to all fields of the AppState. 
Restructure the appstate fields to make a clearer distinction between backend fields/utils and features.

### Motivation

Categorizing code according to features for clarity and redux pattern implementation.

### To-do List

- [x] Impleement the FeatTrait on all fields of the appstate
- [x] restructure fields according to features
- [x] add docs

PR #138, opened 2024-02-26
refactor: refactor `AppState`, add `FeatTrait`

Oran-Dan:
# Description of change
I have changed the appstate struct to fit the redux pattern and the coming folder structure better.

#137 

Since it's a refactoring, the existing tests have been used.


identity-wallet/src/lib.rs
```diff
@@ -5,21 +5,21 @@ pub mod state;
...
-/// The error.rs defines our app_error types, implemented throughout the code using the thiserror crate.
-/// The persistence.rs is where we define our app persistence functions.
-/// The stronghold.rs is where we implement the stronghold library for our app, which is used to store sensitive data.
 // Re-exports
-pub use oid4vc::{oid4vc_core, oid4vc_manager, oid4vci, oid4vp, siopv2};
-
 use jsonwebtoken::{decode, Algorithm, DecodingKey, Validation};
 use lazy_static::lazy_static;
 use log::info;
+pub use oid4vc::{oid4vc_core, oid4vc_manager, oid4vci, oid4vp, siopv2};
```
nanderstabel: Only `pub use oid4vc::{oid4vc_core, oid4vc_manager, oid4vci, oid4vp, siopv2};` are re-exports.

identity-wallet/src/state/actions/mod.rs
```diff
@@ -31,12 +31,36 @@ pub use update_credential_metadata::*;
...
+///
+/// pub async fn new_reducer(state: AppState, action: Action) -> Result<AppState, AppError> {
+///  -- your code -- {
+///      return ( AppState {
+///                  the changes,
+///                  ..state     // Sets the rest to the same values as the old state.
+///             })
+///      }
+///  Ok(state)        // Returns the old state if the state update fails.
+/// }
```
nanderstabel: This example code would not compile and would give warnings for unreachable code and field names not being snake_case (ignoring the pseudocode parts of it).

I think it would be better to make sure that when un-commenting all this code in this example, that everything would actually work, e.g.:
```rust
#[derive(Debug, Serialize, Deserialize, Clone)]
pub struct TestExampleAction {
    example_field: String,
    example_field_2: bool,
}

#[typetag::serde(name = "[Example] Example Action")]
impl ActionTrait for TestExampleAction {
        vec![reducer!(test_example_action)]
    }
}

pub async fn test_example_action(state: AppState, action: Action) -> Result<AppState, AppError> {
    if let Some(test_example_action) = listen::<TestExampleAction>(action) {
        // Reducer logic goes here
        return Ok(AppState {
            // Add changes to the state here
            ..state
        });
    }

    Ok(state)
}
```

nanderstabel: ps: I also renamed the action to `TestExampleAction` to follow the Redux recommended naming convention for Actions :) : https://redux.js.org/style-guide/#write-meaningful-action-names

identity-wallet/src/state/mod.rs
```diff
@@ -56,53 +56,52 @@ pub struct Managers {
 #[ts(export)]
 #[serde(default)]
 pub struct AppState {
+    /// This field contains the connections, containing the useable info for the frontend.
+    pub connections: Vec<Connection>,
+    /// This field contains the display credentials, containing the useable info for the frontend.
```
nanderstabel: I think the `, containing the useable info for the frontend.` can be dropped.

identity-wallet/src/state/mod.rs
```diff
@@ -56,53 +56,52 @@ pub struct Managers {
...
+    pub dev_mode: DevMode,
 }
 
 impl Clone for AppState {
     fn clone(&self) -> Self {
         Self {
-            managers: self.managers.clone(),
-            active_profile: self.active_profile.clone(),
-            active_connection_request: serde_json::from_value(serde_json::json!(self.active_connection_request))
+            back_end_utils: BackEndUtils {
```
nanderstabel: So Tauri makes the distinction between the frontend and Tauri Core (instead of Backend), so we should probably start doing the same.

My suggestions for an alternative name would be `SystemState` or `CoreState`.

identity-wallet/src/state/mod.rs
```diff
@@ -195,25 +243,30 @@ mod tests {
...
+                  "debug_messages": [],
+                  "extensions": {},
+                  "dev_mode": "Off"
                 }"#}
         );
     }
 }
+
+// left: "{\n  \"connections\": [],\n  \"credentials\": [],\n  \"user_data_query\": [],\n  \"profile_settings\": {\n    \"locale\": \"en-US\",\n    \"profile\": {\n      \"name\": \"John Doe\",\n      \"picture\": null,\n      \"theme\": null,\n      \"primary_did\": \"did:example:123\"\n    }\n  },\n  \"current_user_prompt\": {\n    \"type\": \"redirect\",\n    \"target\": \"me\"\n  },\n  \"user_journey\": null,\n  \"debug_messages\": [],\n  \"extensions\": {},\n  \"dev_mode\": \"Off\"\n}"
+// right: "{\n\"connections\": [],\n\"credentials\": [],\n\"user_data_query\": [],\n\"profile_settings\": {\n  \"locale\": \"en-US\",\n  \"profile\": {\n    \"name\": \"John Doe\",\n    \"picture\": null,\n    \"theme\": null,\n    \"primary_did\": \"did:example:123\"\n  }\n},\n\"current_user_prompt\": {\n  \"type\": \"redirect\",\n  \"target\": \"me\"\n},\n\"user_journey\": null,\n\"debug_messages\": [],\n\"extensions\": {},\n\"dev_mode\": \"Off\"\n}"
```
nanderstabel: Can be removed :smiley:

identity-wallet/src/state/persistence.rs
```diff
@@ -22,6 +23,10 @@ pub async fn load_state() -> anyhow::Result<AppState> {
 pub async fn save_state(app_state: &AppState) -> anyhow::Result<()> {
     let state_file = STATE_FILE.lock().unwrap().clone();
     let mut file = File::create(state_file).await?;
+
+    let mut json_app_state = serde_json::to_value(app_state)?;
+    json_app_state["credentials"] = serde_json::Value::Array(Vec::new());
+
```
nanderstabel: Could use a comment explaining why `credential` is persisted here. :+1:

nanderstabel (changes requested):
Great work :smile:

identity-wallet/src/state/persistence.rs
nanderstabel: *is not persisted

Norlock: Maybe you can clone app_state, then remove credentials on rust side before storing. That way you let Rust and serde do the work.

Norlock (commented):
Overal looks good, maybe you can take a look at that one comment

identity-wallet/src/state/persistence.rs
nanderstabel: I guess perhaps each Feature should have the ability to set its persistence strategy, either pure json file or stronghold. So `pub async fn save_state(app_state: &AppState)` would be `pub async fn save_state(app_state: Box<dyn FeatState>)` or something similar.

Perhaps even better it could/should be implemented via the `FeatTrait`. (for another time though, I think the current solution, although not super clean, simply does the job).

nanderstabel (approved):

Norlock (approved):

Issue #62, opened 2023-10-03
Add History


JelleMillenaar:
### Description

Add History feature into the wallet which showcases the following different events: Connecting, Sharing Data and Receiving Data. The events are connected to the Connection and should display the correct logo and display name or website. If not implemented yet, the events must first be tracked and stored. We should consider secure storage vs fast storage for these events. 

### Motivation

While we doubt the feature will be used significantly, it is a vital tool and protection against over-requesting data by connections. The history can give insight into abuse and may later be extended with reporting logic. In addition, it might be useful for users to view their history of events for a range of reasons.

### Resources

Main History Page:
![image](https://github.com/impierce/identity-wallet/assets/29886335/5c9fbb07-dc0f-4511-9428-7c395678fe56)

Connection Specific History:
![image](https://github.com/impierce/identity-wallet/assets/29886335/bc01b74d-ffc3-4d1b-96c7-ee5526ec6c37)

### To-do List

- [ ] Track Events and Store them (Securely or efficiently?) - Care to include signed request?
- [ ] Implement History overview
- [ ] Implement filtered History overview by Connection
- [ ] Support search / filter / order logic based on UX discussion

PR #139, opened 2024-02-26
feat: introduce history events

Norlock:
# Description of change
Please write a summary of your changes and why you made them.

fixes issue #62.
Is based upon: https://github.com/impierce/identity-wallet/pull/122

I ran the Shenron profile and made sure the events where added in the general profile and connection specific.


.github/workflows/format-lint-test.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v4
       with:
-        node-version: '20.x'
+        node-version: '21.x'
```
daniel-mader: Any specific reason for this change? `20.x` is the current LTS, no?

unime/src/lib/components/molecules/ListItemCard.svelte
```diff
@@ -10,6 +10,15 @@
   export let description: string | undefined = undefined;
   export let type: 'data' | 'badge' = 'data';
   export let isTempAsset = false;
+  export let hasBorderRadius = true;
+
+  function addRoundedBorder() {
```
daniel-mader: Please use `twMerge()` and `$$props.class` to extend a component. An example can be found in `ButtonRounded.svelte`.

unime/src-tauri/tests/fixtures/states/credential_redirect_me.json
```diff
@@ -8,7 +8,7 @@
     "credentials": [
         {
             "id": "39373933-3863-3339-3864-646234373631",
-            "issuer_name": null,
+            "issuer_name": "",
```
daniel-mader, line 13: Is an empty string really better than `null`? Null can be more easily checked against and an empty string would be valid, but not display anything. Is this desired behavior?

unime/src/i18n/de-DE/index.ts
```diff
@@ -255,6 +255,9 @@ const de = {
   },
   TIMELINE: {
     EMPTY: 'Noch keine Aktivitäten.',
+    ADDED_CREDENTIALS: 'TODO',
```
daniel-mader: Let's not do this as such things are easily forgotten compared to a type error when using the **_typesafe_**-i18n library. Touching a translation file without adding the actual translations should be avoided.

unime/src/i18n/i18n-types.ts
```diff
@@ -1463,6 +1475,18 @@ export type TranslationFunctions = {
 		 * No activity yet.
 		 */
 		EMPTY: () => LocalizedString
+		/**
+		 * Received credentials from
+		 */
+		ADDED_CREDENTIALS: () => LocalizedString
```
daniel-mader: Have you also translated these new types to `en` and `nl-NL`?

identity-wallet/src/state/reducers/history.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub issuer_name: String,
+    pub event_type: EventType,
+    pub connection_id: Option<String>,
+    pub date: String,
+    pub credentials: Vec<HistoryCredential>,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, TS)]
+#[ts(export)]
+pub struct HistoryCredential {
```
daniel-mader: Do you think it makes sense to add the credential id as field so we could directly jump to that credential when scrolling through the history?

identity-wallet/src/command.rs
```diff
@@ -102,6 +102,9 @@ pub async fn handle_action<R: tauri::Runtime>(
 pub fn emit_event<R: tauri::Runtime>(window: &tauri::Window<R>, app_state: &AppState) -> anyhow::Result<()> {
     const STATE_CHANGED_EVENT: &str = "state-changed";
     window.emit(STATE_CHANGED_EVENT, app_state)?;
-    debug!("emitted event `{}` with payload `{:?}`", STATE_CHANGED_EVENT, app_state);
+
+    let json_str = serde_json::to_string(app_state).unwrap();
```
daniel-mader: I suggest either inlining into `debug!()` or give it a proper name such as `app_state_json_string`.

identity-wallet/src/state/reducers/history.rs
```diff
@@ -0,0 +1,31 @@
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, TS)]
+#[ts(export)]
+pub enum EventType {
+    #[serde(rename = "added_connection")]
+    AddedConnection,
```
daniel-mader: The [naming convention for events](https://www.eventstore.com/blog/whats-in-an-event-name#naming-conventions) is **SubjectObjectPredicate** so: `AddedConnection` --> `ConnectionAdded`

.github/workflows/format-lint-test.yaml
Norlock: I had problems with format not working locally so I changed this one, (eventually it was tsconfig.json that was formatted incorrectly so I added it to the prettierignore). I can revert this change I thought it was fine because it still worked and was newer

unime/src-tauri/tests/fixtures/states/credential_redirect_me.json
Norlock: Yes I also discussed with @nanderstabel  that issuer_name is always set in the code, so it can never be null anyway. We need to make a new PR in the future where we look at the frontend ?? checks on the issuer_name to incorporate those back to the backend again.

identity-wallet/src/state/reducers/history.rs
Norlock: Hmm yes seems like a good idea, I will add the field to it.

identity-wallet/src/state/reducers/history.rs
Norlock: Ah ok I will rename :+1:

unime/src/i18n/de-DE/index.ts
Norlock: ok no problem, I do think we should aim for strict typescript compiling, which means that no errors can exist in typing. Which would make this solution not possible, but for now I remove the fields again

unime/src/lib/components/molecules/ListItemCard.svelte
Norlock: But I want to always use the same class, except for in history I don't want to add a certain class. Isn't it better to do it like this? Otherwise everywhere you have to pass rounded-lg class

.github/workflows/format-lint-test.yaml
Norlock: changed it back to 20.x

unime/src/i18n/de-DE/index.ts
daniel-mader: Yes, but throwing type errors when translations are incomplete is the main idea behind **_typesafe_**-i18n. Keeping all translations up-to-date at all time for all features is a challenge. Maybe we find an elegant solution when it's the right time. 👍

identity-wallet/src/state/reducers/user_data_query.rs
daniel-mader, line 1: I'm not questioning these improvements, but is it really necessary to clean up `user_data_query.rs` in the same PR that introduces history events? I know it's just "small issues", but they still delay the review process and bloat the initial scope.

identity-wallet/src/state/user_prompt.rs
```diff
@@ -33,9 +33,9 @@ pub enum CurrentUserPrompt {
         issuer_name: String,
         #[ts(optional)]
         logo_uri: Option<String>,
-        #[ts(type = "object")]
+        #[ts(type = "any")]
```
daniel-mader, line 36: Is this related to history events?

identity-wallet/src/verifiable_credential_record.rs
```diff
@@ -68,18 +74,40 @@ impl From<CredentialFormats<WithCredential>> for VerifiableCredentialRecord {
...
+        Value::Array(array) => array.last()?.as_str().map(|name| name.to_string()),
+        _ => None,
+    }
+}
+
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, Derivative, TS)]
 #[derivative(PartialEq)]
 #[ts(export, export_to = "bindings/display-credential/DisplayCredential.ts")]
 pub struct DisplayCredential {
+    /// Also display_icon
```
daniel-mader: I don't think we need to comment that the `id` is also used somewhere else as it's not relevant for the `DisplayCredential`. Also, it's not "display_icon", rather "image id".

unime/src/lib/credentials/CredentialList.svelte
```diff
@@ -98,9 +98,7 @@
     {#each credentials as credential}
       <ListItemCard
         id={credential.id}
-        title={credential.metadata.display.name ??
-          credential.data.credentialSubject.achievement?.name ??
-          credential.data.type.at(-1)}
+        title={credential.display_name}
```
daniel-mader, line 101: Are we still making sure that there is _something_ displayed in any case? If `display_name` can be an empty string, then we won't see anything in the UI, right?

unime/src/routes/(app)/activity/connection/[id]/+page@.svelte
```diff
@@ -30,9 +27,9 @@
         <ConnectionData id={connection.id} />
       </div>
 
-      <div slot="2" class="h-full bg-silver py-5 dark:bg-navy">
+      <div slot="2" class="bg-silver pt-5 dark:bg-navy">
```
Comment (daniel-mader) [inline unime/src/routes/(app)/activity/connection/[id]/+page@.svelte:30]: Why remove the padding from the bottom?

unime/src/routes/(app)/activity/+page.svelte
```diff
@@ -24,7 +24,7 @@
       </button>
     </div> -->
   </div>
-  <div class="flex grow flex-col overflow-y-auto px-4 py-5">
+  <div class="flex grow flex-col overflow-y-auto px-4 pt-5">
```
daniel-mader, line 27: Why remove the padding from the bottom?

daniel-mader: Okay, overflows are looking nicer 👍

unime/.prettierignore
```diff
@@ -5,6 +5,7 @@ node_modules
 /package
 .env
 .env.*
+.tsconfig.json
```
daniel-mader: Did you have problems with `tsconfig` being formatted differently on your machine?

identity-wallet/src/state/reducers/history.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub date: String,
+    pub credentials: Vec<HistoryCredential>,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, TS)]
+#[ts(export)]
+pub struct HistoryCredential {
+    pub title: String,
+    pub sub_title: String,
+    /// Is also the image id
```
daniel-mader: I don't think we need to be aware here that this `id` is used elsewhere. You can safely remove this comment.

identity-wallet/src/state/reducers/history.rs
```diff
@@ -0,0 +1,32 @@
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, TS)]
+#[ts(export)]
+pub enum EventType {
+    #[serde(rename = "connection_added")]
```
daniel-mader: Why rename? `"ConnectionAdded"` is perfectly idiomatic for the frontend as well.

identity-wallet/bindings/EventType.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+export type EventType = "connection_added" | "credentials_added" | "credentials_shared";
```
daniel-mader: I think `"ConnectionAdded"` is more conventional than `"connection_added"` for the frontend.

identity-wallet/src/state/user_prompt.rs
Norlock: Its part of the other PR where this one was built upon.

identity-wallet/src/state/reducers/user_data_query.rs
Norlock: It was needed because display_name has been set in the backend and is not an optional value anymore. ALso part of the previous PR

identity-wallet/src/verifiable_credential_record.rs
Norlock: Well it just makes everything easier. Because last time we took a long time finding out where to find the "image id" and we had to look through the code until you realized it was that field.

unime/.prettierignore
Norlock: Yes it adds trailing slashes on that file, which will fail again on github actions

unime/src/lib/credentials/CredentialList.svelte
Norlock: Yes because this is being added in the backend. Its part of the previous PR, we talked about few weeks ago where I said I would move this to the backend.

identity-wallet/src/verifiable_credential_record.rs
daniel-mader: Let's make a proper comment out of it then such as "This id is also used as the image asset id"

unime/src/lib/credentials/CredentialList.svelte
daniel-mader: Maybe I commented in the wrong place (frontend), but would we handle `display_name` to never be an empty string?

Norlock: Well it can be an empty string but the whole idea of a display_name is that it's always set right? I think we should require that some fields are always set. Because to have a name missing of a credential in my opinion is not a correct state. Because to the end user it's hard to make sense of what the credential is otherwise right

daniel-mader: Yes, correct. That's what I was asking. We should make sure it can never be empty in the backend. We do that, right?

Norlock: I just moved those three ?? fields to the backend. If something is still missing it would have been null or empty before as well, so maybe for another story. To see where its added now: identity-wallet/src/verifiable_credential_record.rs

nanderstabel:
Currently when I connect with the same Connection multiple times, they will all show up as a "First connection". I believe from the 
#62 after the first connection we should refer to 're-connecting' as simply "Connected to ...".

Screenshot_20240229-175705

unime/src/lib/events/History.svelte
```diff
@@ -1,71 +1,99 @@
...
+  import type { HistoryEvent } from '@bindings/HistoryEvent';
+
   import HistoryEntry from '$lib/events/HistoryEntry.svelte';
   import LL from '$src/i18n/i18n-svelte';
-  import exampleEvents from '$src/lib/events/mock-data.json';
   import { state } from '$src/stores';
 
-  // const events: Event[] = exampleEvents.map((e) => ({ ...e, type: e.type as EventType }));
-  const events: Event[] = [];
+  import DownloadSimple from '~icons/ph/download-simple';
```
daniel-mader: I think we're using the "fill" version so you could use:

```
import DownloadSimple from '~icons/ph/download-simple-fill';
import PlugsConnected from '~icons/ph/plugs-connected-fill';
import ShareFat from '~icons/ph/share-fat-fill';
```

unime/src/lib/events/History.svelte
```diff
@@ -1,71 +1,99 @@
...
+              <svelte:component this={event.icon} class="h-4 w-4 " />
+            </div>
+            {#if hasNextElement(i)}
+              <!-- Vertical line on the left -->
+              <div class="mb-2 mt-4 h-full rounded-full border border-slate-200 border-y-gray-200"></div>
+            {/if}
           </div>
-          <div class="grow">
-            <HistoryEntry {...event} />
+          <div class="ml-6 mt-[-8px] flex grow justify-between pb-10">
```
daniel-mader: If you want to align it in the middle, you'd need `mt-[-5px]` since two lines of text are `42px` and the icon to left is `32px`, so `(42px - 32px) / 2 = 5px`.

daniel-mader: Using a non-absolute way of alignment would be better (using flex), but it's okay for now.

unime/src/lib/events/History.svelte
```diff
@@ -1,71 +1,99 @@
...
+    <div class="ml-2 mt-6">
+      {#each events as event, i}
+        <div class="flex flex-row">
+          <div class="mt-2 flex flex-col items-center">
+            <div class="z-10 flex items-center justify-center rounded-full bg-white ring-8 ring-white dark:ring-silver">
+              <svelte:component this={event.icon} class="h-4 w-4 " />
+            </div>
+            {#if hasNextElement(i)}
+              <!-- Vertical line on the left -->
+              <div class="mb-2 mt-4 h-full rounded-full border border-slate-200 border-y-gray-200"></div>
```
daniel-mader: Using borders here is not necessary, you can just use a regular width, such as `w-[2px]` and remove all borders. Also, please add `bg-grey dark:bg-blue`.

unime/src/lib/events/History.svelte
```diff
@@ -1,71 +1,99 @@
...
-    <div class="flex grow flex-col space-y-8 pr-4 pt-4">
-      {#each eventsList as event}
-        <div class="flex justify-between">
-          <div class="z-10 mr-3 h-6 w-6 overflow-hidden rounded-full bg-white p-0.5 ring-8 ring-silver">
-            <Image id={event.connection.id} imgClass="h-full w-full" />
+    <div class="ml-2 mt-6">
+      {#each events as event, i}
+        <div class="flex flex-row">
+          <div class="mt-2 flex flex-col items-center">
+            <div class="z-10 flex items-center justify-center rounded-full bg-white ring-8 ring-white dark:ring-silver">
```
daniel-mader:
```suggestion
            <div class="z-10 flex items-center justify-center rounded-full bg-grey text-slate-800 ring-8 ring-grey dark:bg-blue dark:text-grey dark:ring-blue">
```

daniel-mader: There colors are a little bit off here in comparison with the UI designs.

identity-wallet/src/state/reducers/authorization.rs
```diff
@@ -261,19 +279,33 @@ pub async fn handle_oid4vp_authorization_request(state: AppState, action: Action
...
+
         let verifiable_credentials: Vec<Credential<JwtVcJson>> = stronghold_manager
             .values()
             .map_err(StrongholdValuesError)?
             .unwrap()
             .iter()
-            .filter_map(
-                |verifiable_credential_record| match &verifiable_credential_record.verifiable_credential {
+            .filter_map(|verifiable_credential_record| {
+                let share_cred = match &verifiable_credential_record.verifiable_credential {
```
nanderstabel: Preferably please follow the general naming convention by avoiding abbreviations :)

identity-wallet/src/state/reducers/credential_offer.rs
```diff
@@ -359,25 +366,40 @@ pub async fn send_credential_request(state: AppState, action: Action) -> Result<
...
             .map(|verifiable_credential_record| verifiable_credential_record.display_credential)
             .collect();
 
-        let current_user_prompt = CurrentUserPrompt::Redirect {
-            target: "me".to_string(),
-        };
+        // History
+        if !history_credentials.is_empty() {
+            state.history.push(HistoryEvent {
+                issuer_name: issuer_name.clone(),
```
nanderstabel: This does not need to be cloned :+1:

unime/src/lib/events/HistoryEntry.svelte
```diff
@@ -1,34 +1,75 @@
...
         <ListItemCard
           id={credential.id}
-          title={credential.metadata.display.name ?? credential.data.type.at(-1)}
-          description="Lorem ipsum dolor"
-        >
-          <!-- <span slot="icon"><EnvelopeSimple class="h-6 w-6" /></span> -->
-        </ListItemCard>
+          title={credential.title}
+          description={credential.sub_title}
+          class="rounded-none"
```
daniel-mader: Why remove the rounded borders?
* It does not have a visual effect since background and element have the same color.
* Also when they don't have the same color, it looks worse without rounded borders.

unime/src/lib/events/HistoryEntry.svelte
```diff
@@ -1,34 +1,75 @@
...
-      dateStyle: 'medium',
-      timeStyle: 'medium',
-    })}
+  <p class="text-[14px]/[22px] font-medium text-slate-800 dark:text-grey">{title}</p>
+  <p class="text-[12px]/[20px] font-medium text-slate-500 dark:text-slate-300">
+    {getRelativeDate()}
   </p>
   {#if credentials.length > 0}
-    <div class="mt-[12px] rounded-xl border border-slate-200 bg-white p-[3px] dark:border-slate-600">
+    <div class="mt-[12px] rounded-xl border border-y-gray-200 bg-white p-[3px] dark:border-dark dark:bg-dark">
```
daniel-mader: * These changes do not reflect the UI designs (they use `slate`, not `gray`).
* Why are only the y-axis borders styled?

identity-wallet/bindings/HistoryCredential.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+export interface HistoryCredential { title: string, sub_title: string, id: string, }
```
daniel-mader: `sub_title` is non-conventional naming for JS/TS. It always contains the `issuerName`, right? So you could also just use that. If you want to keep it more generic, you could also name it `description` (as in the generic `ListItemCard`).

unime/src/lib/events/History.svelte
Norlock: I only see light theme design? Where did you see the dark theme design?

identity-wallet/src/state/reducers/history.rs
```diff
@@ -0,0 +1,29 @@
...
+pub enum EventType {
+    ConnectionAdded,
+    CredentialsAdded,
+    CredentialsShared,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, TS)]
+#[ts(export)]
+pub struct HistoryEvent {
+    pub issuer_name: String,
```
nanderstabel: Probably better to change this to a more generic name, such as `name` or `connection_name`. 

In most cases issuers and verifiers are very distinct entities so `issuer_name` is not the right term when credentials are shared.

unime/src/lib/events/History.svelte
daniel-mader: All features are designed only in light mode and we have to apply the dark theme ourselves by following the color mapping.

unime/src/i18n/en/index.ts
```diff
@@ -254,6 +254,9 @@ const en = {
   },
   TIMELINE: {
     EMPTY: 'No activity yet.',
+    CREDENTIALS_ADDED: 'Received credentials from',
+    CREDENTIALS_SHARED: 'Shared credentials with',
```
daniel-mader: Isn't it "Shared **data** with" instead of "Shared **credentials** with"?

daniel-mader: Same for "Received **data** from", right?

unime/src/lib/components/molecules/ListItemCard.svelte
```diff
@@ -1,6 +1,8 @@
 <script lang="ts">
   import { createEventDispatcher } from 'svelte';
 
+  import { twMerge } from 'tailwind-merge';
```
daniel-mader: Unused import

identity-wallet/src/state/reducers/history.rs
Oran-Dan, line 1: These structs are in fact not reducers, so they are in the wrong folder. Anticipating the folder restructure, you can create a shared folder and add them to a separate shared/history.rs file. This will be the required structure when your branch gets merged. It will still create a conflict but then you won't have to change the imports anymore after merging.

Oran-Dan:
I suggest we use only one name for the dragon profile.
A bit confusing the Shenron and dragon double naming. 
Since all functions and enums use Dragon i suggest we just use Dragon 100% also as name and leave the Shenron.
Sorry for commenting this way too late on a wrong PR :|

unime/src/i18n/en/index.ts
Norlock: Yes but I discussed with Andrew and he seems to prefer credentials as well, however if more than just credentials will be shared we can use data. Its more used as an umbrella term which could include more.

unime/src/i18n/en/index.ts
```diff
@@ -207,7 +207,7 @@ const en = {
     NAVBAR_TITLE: 'Connected',
     TABS: {
       CONNECTIONS: 'Connections',
-      TIMELINE: 'Timeline',
+      HISTORY: 'History',
```
nanderstabel, line 210: Can you update these tags for NL and DE as well? (Same for `HISTORY` below in line 255)

nanderstabel (dismissed):

Oran-Dan (dismissed):

unime/.prettierignore
daniel-mader: The file `.tsconfig.json` does not exist. You might mean `tsconfig.json`, but I suggest you undo that ignore and reintroduce it if you encounter issues again.

nanderstabel (approved):

daniel-mader (approved):

Issue #125, opened 2024-02-22
Treat Credential Offers and OID4VP Authorization Requests as connections


nanderstabel:
### Description

Currently only when  a SIOPv2 authorization request is accepted, the relying party is added as a connection.
We need to make sure that Credential Issuers are treated as a Connection as well when a Credential Offer is accepted.
Also when an OID4VP Authorization Request is accepted for an unknown Relying Party, this needs to be treated as a new Connection as well.

### Motivation

Connections should not be limited to only Relying Parties that request for authorization. Any interaction with either a Relying Party or Credential Issuer should be considered a new (interaction with a) Connection. 

### Resources

Related to #62 
Related to #155

### To-do List

- [ ] implement Credential Offers as Connections
- [ ] implement OID4VP Authorization Requests as Connections

PR #151, opened 2024-03-12
fix: derive `connection_id` from `client_name` and make it non-optional

### LINKED ISSUE #150: [Bug] Not all events shown through the connection view
### Description

When you click on a connection an event is missing.

### Hardware Specification

_No response_

1. Run the Shenron profile
2. click on the added connection
4. Go to the history of that connection (missing the received credential)

### Expected Behaviour

To see all history events related to that issuer

### Actual Behaviour

It misses one history event

### Errors
Connection history without the future engineer:
image (1)

Overal history (with the future engineer):
image (2)

Norlock:
# Description of change
Connection history didn't show all events, because some were wrongly filtered out. This also fix a bug where date wasn't correctly when receiving credentials.

Fixes #150

I clicked on the connection and looked again

### DIFF
--- a/identity-wallet/src/date_utils.rs
+++ b/identity-wallet/src/date_utils.rs
@@ -1,0 +1,7 @@
+pub struct DateUtils;
+
+impl DateUtils {
+    pub fn new_date_string() -> String {
+        chrono::Utc::now().to_rfc3339()
+    }
+}
--- a/identity-wallet/src/date_utils.rs
+++ b/identity-wallet/src/date_utils.rs
@@ -1,7 +1,0 @@
-pub struct DateUtils;
-
-impl DateUtils {
-    pub fn new_date_string() -> String {
-        chrono::Utc::now().to_rfc3339()
-    }
-}
--- a/identity-wallet/src/state/connections/mod.rs
+++ b/identity-wallet/src/state/connections/mod.rs
@@ -21,3 +21,12 @@ pub struct Connection {
 
 #[typetag::serde(name = "connection")]
 impl FeatTrait for Connection {}
+
+impl Connection {
+    // TODO: This is a HORRIBLE solution to determine the connection_id by the non-unique "issuer name".
+    // It is a TEMPORARY solution and should only be used in DEMO environments,
+    // since we currently lack a unique identitfier to distinguish connections.
+    pub fn create_connection_id(client_name: &str) -> String {
+        base64::encode_config(&client_name, base64::URL_SAFE)
+    }
+}
--- a/identity-wallet/src/state/connections/mod.rs
+++ b/identity-wallet/src/state/connections/mod.rs
@@ -23,9 +23,7 @@ pub struct Connection {
 impl FeatTrait for Connection {}
 
 impl Connection {
-    // TODO: This is a HORRIBLE solution to determine the connection_id by the non-unique "issuer name".
-    // It is a TEMPORARY solution and should only be used in DEMO environments,
-    // since we currently lack a unique identitfier to distinguish connections.
+    // TODO(ngdil): Temporary solution to support NGDIL demo, replace with different unique identifier to distinguish connection
     pub fn create_connection_id(client_name: &str) -> String {
         base64::encode_config(&client_name, base64::URL_SAFE)
     }
--- a/identity-wallet/src/state/connections/mod.rs
+++ b/identity-wallet/src/state/connections/mod.rs
@@ -24,7 +24,7 @@ impl FeatTrait for Connection {}
 
 impl Connection {
     // TODO(ngdil): Temporary solution to support NGDIL demo, replace with different unique identifier to distinguish connection
-    pub fn create_connection_id(client_name: &str) -> String {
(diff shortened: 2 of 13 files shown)


Oran-Dan (dismissed):

identity-wallet/src/state/connections/reducers/handle_siopv2_authorization_request.rs
```diff
@@ -64,10 +64,7 @@ pub async fn handle_siopv2_authorization_request(mut state: AppState, _action: A
             connection.last_interacted = connection_time.clone();
         });
 
-    // TODO: This is a HORRIBLE solution to determine the connection_id by the non-unique "issuer name".
-    // It is a TEMPORARY solution and should only be used in DEMO environments,
-    // since we currently lack a unique identitfier to distinguish connections.
-    let connection_id = base64::encode_config(&client_name, base64::URL_SAFE);
```
nanderstabel, line 70: `handle_oid4vp_authorization_request.rs` has the same TODO message so It would be nice to replace that there as well :)

identity-wallet/src/state/connections/mod.rs
```diff
@@ -21,3 +21,12 @@ pub struct Connection {
 
 #[typetag::serde(name = "connection")]
 impl FeatTrait for Connection {}
+
+impl Connection {
+    // TODO: This is a HORRIBLE solution to determine the connection_id by the non-unique "issuer name".
+    // It is a TEMPORARY solution and should only be used in DEMO environments,
+    // since we currently lack a unique identitfier to distinguish connections.
```
nanderstabel: I think this is a nice opportunity to replace this somewhat screamy TODO message to :
```
// TODO(ngdil): Temporary solution to support NGDIL demo, replace with different unique identifier to distinguish connection
```

identity-wallet/src/state/core_utils/history_event.rs
```diff
@@ -14,7 +14,7 @@ pub enum EventType {
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq, TS)]
 #[ts(export)]
 pub struct HistoryEvent {
-    pub connection_id: Option<String>,
+    pub connection_id: String,
```
nanderstabel, line 17: Since this type has changed, we should also update the bindings :+1: 

Do you think there might also be some changes needed in the frontend once the binding is updated?

nanderstabel (changes requested):
Nice fixes! I think it's good to adjust the PR title and description a bit so that it also reflects the second fix you made now.

Also it the code needs a `cargo fmt` :+1: 

~~The second fix fixes issue  #125 btw so you can also add that to the description :smiley:~~

identity-wallet/src/state/core_utils/history_event.rs
Norlock: No its not needed luckily

identity-wallet/src/state/connections/mod.rs
```diff
@@ -21,3 +21,10 @@ pub struct Connection {
 
 #[typetag::serde(name = "connection")]
 impl FeatTrait for Connection {}
+
+impl Connection {
+    // TODO(ngdil): Temporary solution to support NGDIL demo, replace with different unique identifier to distinguish connection
+    pub fn create_connection_id(client_name: &str) -> String {
+        base64::encode_config(&client_name, base64::URL_SAFE)
```
nanderstabel: We now use this method for both `client_name`'s and `issuer_name`'s. To avoid confusion I think it's slightly better to go for the more generic `connection_name` as variable name her (instead of `client_name`).

nanderstabel (dismissed):

daniel-mader (commented):
Is there any code changes required to fulfill the "treat credential offers as connections"?

identity-wallet/src/date_utils.rs
Oran-Dan, line 1: Can you explain the reasoning behind this? it seems a bit bloaty to me?
In anyway, it should be in the core_utils folder, but i don't see the use of putting this into a file all together?
a line saying chrono::Utc::now().to-rfc3339() in the code is fine too right?

identity-wallet/src/state/credentials/mod.rs
Oran-Dan, line 1: For example here, i don't see the gain of adding a struct, impl and a new file to the codebase? 
What was wrong with the old version?

Norlock: It is now used in multiple places, so to prevent inconsistencies whenever something changes, I made it a global function. Yes its a bit much too have a new file, but I think it doesn't fit in any existing file if it supposed to be a global function. 

The reason for a struct is that it implies contexts. If you have a lot of global functions without a struct you can get name conflicts quicker and you need to search for the function name. Now you can add DateUtils and then check the function you want from it. 

I will move it to core_utils :+1:

daniel-mader (dismissed):

Oran-Dan (approved):

daniel-mader (approved):

PR #153, opened 2024-03-13
fix: make theme non-optional (default: `system`)

### LINKED ISSUE #152: [Bug] Theme behavior inconsistent during onboarding
### Description

During onboarding I get the light theme and it switches to dark (system) after onboarding.
The same on me laptop as on my iphone testflight app.
When trying to reproduce the problem it (strangely) seems to cache the previous theme anyway on my testflight as well as laptop. After a rebuild on my laptop it is white again though until the above mentioned bug appears again.

Maybe an easy fix to make this consistent since we just added default values for the name and picture as well ("Me" and the initials). Maybe we can fill in those default values even before onboarding already, so that the profile is set and the theme won't act erratic. --> profile (name:"Me", picture:"", theme:"system", primary_did: ?? will this field make a problem?)

### Hardware Specification

Operating system: Linux ubuntu 22.04 & iOS 16.7.1
RAM: 
Cores:
Device: Laptop HP elitebook  & iPhone X

1. Fresh rebuild or download and the onboarding will be in the light theme until it's done and automatically switches to system (which is almost always dark on my devices)
2. Delete profile and you will see it seems to save the last selected theme someway for the new onboarding, so this can be light/dark/system depending on what you selected before deleting your profile.
3.

### Expected Behaviour

it shouldnt unrequestedly switch theme after onboarding, it should be system theme from the very start.

### Actual Behaviour

it should be system theme from the start

### Errors

no errors

Norlock:
# Description of change
Make theme default (system) because it would be that. However the bug was related to when profile was not set, it wouldn't go into the function because it would be null (see the if guard). However this fix makes sure it will default back to system 

#152 .

Created profiles with system, dark and light, and then reset the profile again.


Oran-Dan (dismissed):

Oran-Dan (dismissed):

nanderstabel:
Something still seems the be not working correctly here (at least on my device).

This scenario:
1. my phone is in dark mode
2. I start UniMe, it picks up on the system theme which is Dark mode, I just stay in the welcome page :white_check_mark:  
3. I don't close the app, but I change my phone phone to Light mode
4. I go back to UniMe, but it is still in Dark mode :x: 

point 4. used to automatically switch theme (it still did in  `0.4.10`)

unime/src/routes/+layout.svelte
```diff
@@ -39,23 +39,14 @@
 
   const systemColorScheme = window.matchMedia('(prefers-color-scheme: dark)');
 
-  systemColorScheme.addEventListener(
```
daniel-mader, line 42: By removing this, you remove the listener to window events completely. When the theme is now changed through the OS, it does not have any effect anymore.

daniel-mader (changes requested):

unime/src/routes/+layout.svelte
```diff
@@ -39,23 +39,24 @@
...
-    'change',
-    (e) => {
-      if ($state?.profile_settings.profile?.theme) {
-        determineTheme(e.matches, $state.profile_settings.profile.theme);
-      }
-    },
-    { once: true },
-  );
+  // This is used for detecting the OS theme changes.
```
daniel-mader: I don't think this comment adds any value since the variable is already called `systemColorScheme`.

unime/src/routes/+layout.svelte
```diff
@@ -39,23 +39,24 @@
...
+      determineTheme(systemColorScheme.matches, 'system');
+    }
+  });
 
   $: {
     // TODO: needs to be called at least once to trigger subscribers --> better way to do this?
     console.log('+layout.svelte: state', $state);
 
-    // needed again?
+    // This is used for detecting the theme changes inside the app.
```
daniel-mader: I don't think this comment adds any value since the check is already descriptive: `$state` > `profile_settings` > `theme`.

daniel-mader (approved):

nanderstabel (approved):

PR #157, opened 2024-03-15
feat: treat Credential Offers and OID4VP Authorization Requests as connections

### LINKED ISSUE #125: Treat Credential Offers and OID4VP Authorization Requests as connections

nanderstabel:
# Description of change
This change ensures that when there is a `CredentialsShared` or a `CredentialsAdded` event with a new Connection, that a `ConnectionAdded` event will be fired first. 

This change also ensures that the `connection_url` can be derived from both the `credential_issuer_url` and the `redirect_url` in a similar fashion. This means that if a server is both a Credential Issuer as well as a Relying Party, then it's `connection_url`  while in the Credential Issuer role will be equal to  it's `connection_url` while in the Relying Party role.

This in turn enables us to use the issuer/client_name together with the `connection_url` to generate Connection IDs (instead of just based on the issuer/client_name which has severe security risks).

Abstracting the code for inserting and updating of Connections results in cleaner and less error-prone code:
```rust
        let previously_connected = state.connections.contains(connection_url, &issuer_name);
        let mut connections = state.connections;
        let connection = connections.insert_or_update(connection_url, &issuer_name);
```

At a later stage we could implement something similar for History, e.g:
```rust

        // History
        if !history_credentials.is_empty() {
            // Only add a `ConnectionAdded` event if the connection was not previously connected.
            if !previously_connected {
                state.history.push(HistoryEvent {
                    connection_name: connection.name.clone(),
                    event_type: EventType::ConnectionAdded,
                    connection_id: connection.id.clone(),
                    date: connection.last_interacted.clone(),
                    credentials: vec![],
                });
            }
            state.history.push(HistoryEvent {
                connection_name: connection.name.clone(),
                event_type: EventType::CredentialsAdded,
                connection_id: connection.id.clone(),
                date: connection.last_interacted.clone(),
                credentials: history_credentials,
            });
        }
```

would be converted to:
```rust

        // History
        if !history_credentials.is_empty() {
            // Only add a `ConnectionAdded` event if the connection was not previously connected.
            if !previously_connected {
                state.history.append_connection_added_event(connection, vec![]);
            }
            state.history.append_credentials_added_event(connection, history_credentials);
        }
```

fixes #125 

* Updated integration test to handle `history` and `connections`.
* Added unit tests for the `Connections` struct.
* Manual testing with both NGDIL and the SURF edubadges demo website


identity-wallet/src/state/connections/mod.rs
```diff
@@ -1,30 +1,181 @@
...
+        // TODO(ngdil): Temporary solution to support NGDIL demo, replace with different unique identifier to distinguish connection
+        let id = base64::encode_config([name.as_bytes(), url.as_bytes()].concat(), base64::URL_SAFE);
+        let current_datetime = DateUtils::new_date_string();
+        Self {
+            id,
+            name,
+            url,
+            verified: false,
+            first_interacted: current_datetime.clone(),
+            last_interacted: current_datetime.clone(),
```
Norlock: doesn't need a clone here

identity-wallet/src/state/core_utils/history_event.rs
```diff
@@ -41,3 +41,13 @@ impl HistoryCredential {
         }
     }
 }
+
+/// Implement PartialEq for HistoryEvent to allow for comparison of HistoryEvent instances in tests.
+impl PartialEq for HistoryEvent {
+    fn eq(&self, other: &Self) -> bool {
+        self.connection_id == other.connection_id
+            && self.connection_name == other.connection_name
+            && self.event_type == other.event_type
```
Norlock, line 51: Maybe remove the partial eq so we can't create bugs in the future (date is missing)

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -124,44 +123,34 @@ pub async fn handle_oid4vp_authorization_request(mut state: AppState, action: Ac
...
-                url: connection_url,
-                verified: false,
-                first_interacted: connection_time.clone(),
-                last_interacted: connection_time.clone(),
-            })
-        };
-
-        state.connections = connections;
+        let previously_connected = state.connections.contains(connection_url.as_str(), &client_name);
+        let mut connections = state.connections;
```
Norlock, line 131: Maybe use `let connections = &mut state.connections;`. 

So you also can't forget to move the value back. (line 153). However it will break the rule of not using mut state, its up to you what you want to do. Also for the other comment

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -124,44 +123,34 @@ pub async fn handle_oid4vp_authorization_request(mut state: AppState, action: Ac
...
+        let connection = connections.insert_or_update(&connection_url, &client_name);
 
         // History
+        if !previously_connected {
+            // Only add a `ConnectionAdded` event if the connection was not previously connected.
+            state.history.push(HistoryEvent {
```
Norlock, line 137: Can we move the history event to the Connections struct. So when we add a connection from anywhere in the code we always know a history event is added as well.

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -176,12 +165,10 @@ pub async fn handle_oid4vp_authorization_request(mut state: AppState, action: Ac
...
 ) -> anyhow::Result<(String, Option<String>, String)> {
-    let connection_url = oid4vp_authorization_request
-        .body
-        .redirect_uri
-        .domain()
-        .ok_or(anyhow::anyhow!("unable to get domain from redirect_uri"))?
-        .to_string();
+    // Get the connection url from the redirect url host (or use the redirect url if it does not
+    // contain a host).
+    let redirect_url = oid4vp_authorization_request.body.redirect_uri.clone();
```
Norlock: maybe name redirect_uri if it is a uri

identity-wallet/src/state/credentials/reducers/send_credential_request.rs
```diff
@@ -198,19 +200,32 @@ pub async fn send_credential_request(mut state: AppState, action: Action) -> Res
             .map(|verifiable_credential_record| verifiable_credential_record.display_credential)
             .collect();
 
+        let previously_connected = state.connections.contains(connection_url, &issuer_name);
+        let mut connections = state.connections;
```
Norlock, line 204: Maybe use &mut for state.connections;

Norlock (changes requested):

identity-wallet/src/state/connections/mod.rs
Oran-Dan, line 1: Not sure if it's a frequent or realistic scenario but just wondering:
So if it's a bigger organization and it wants to connect with the name url but under a different name, let's say a school with subdivisions. For example:
previously connected: url = university.com & name = bachelor Math
new connection: url = university.com & name = bachelor Law

So we have previously connected to this url but not with this name.
How do we handle this? Are we satisfied with the current way of handling which is simply showing it as a new connection if I understand correctly.
Because if my train if thought is correct, a university might opt for specifying a different name for each of it's subdomains and depending on how many credentials you get (i've heard talk of a credential per course).
This could lead to some messy situations connection-wise

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: This is actually a good example of why I like the `state` to be immutable by default :) 

My preferred impementation is to go back to how the reducers where initially designed, so in this reducer it would be like this:
```rust
       let previously_connected = state.connections.contains(connection_url.as_str(), &client_name);
        let mut connections = state.connections; // <-- I only want this to be mutable.
        let connection = connections.insert_or_update(&connection_url, &client_name);

        let mut history = state.history; // <-- as well as this
        // It is very easy to see what data I want to change in this reducer. All other fields have no business to be mutable so always safer to keep them immutable by default

        // History
        if !previously_connected {
            // Only add a `ConnectionAdded` event if the connection was not previously connected.
            history.push(HistoryEvent {
                connection_name: connection.name.clone(),
                event_type: EventType::ConnectionAdded,
                connection_id: connection.id.clone(),
                date: connection.last_interacted.clone(),
                credentials: vec![],
            });
        }
        history.push(HistoryEvent {
            connection_name: connection.name.clone(),
            event_type: EventType::CredentialsShared,
            connection_id: connection.id.clone(),
            date: connection.last_interacted.clone(),
            credentials: history_credentials,
        });

        drop(state_guard);
        // Now down here, again it is very easy to see what parts of the state will be updated.
        return Ok(AppState {
            connections,
            history,
            current_user_prompt: Some(CurrentUserPrompt::Redirect {
                target: "me".to_string(),
            }),
            ..state
        });
```

Now the added benefit of using this approach, is that the rust compiler gives me a nice compile error when I forgot to add one of the data variables that I have been mutating back to the `AppState` that will be returned:
```rust
        return Ok(AppState {
            // connections, <-- forgot to put this back into the state
            history,
            current_user_prompt: Some(CurrentUserPrompt::Redirect {
                target: "me".to_string(),
            }),
            ..state
        });
```
will result in this error:
```
131 |           let mut connections = state.connections;
    |                                 ----------------- value moved here
...
156 |           return Ok(AppState {
    |  ___________________^
157 | |             // connections,
158 | |             history,
159 | |             current_user_prompt: Some(CurrentUserPrompt::Redirect {
...   |
162 | |             ..state
163 | |         });
    | |_________^ value used here after move
```

So Rust is trying to take ownership of everything in the current state like this `..state`, but it is failing because I explicitly moved ownership over that field in line 131.

This way I will always be reminded to make sure to add all mutated data in the updated `AppState` that will be returned :)

identity-wallet/src/state/core_utils/history_event.rs
nanderstabel: This `PartialEq` as well as the `PartialEq` impl for `Connection` are solely for testing use-cases. We currently don't need this implementation elsewhere and I assume it is unlikely that we will need it in the near future.

But it is a bit confusing like this, so eventually it will be cleaner to put these `impl`s behind some "test" feature flag. For now I added a TODO to make it a bit more clear :+1:

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: I had a similar idea of abstracting the `HistoryEvents` that I added to this PR description. We can definitely do something like you suggest although I think it's slightly out of the scope of this PR

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: good catch, thanks :+1:

identity-wallet/src/state/credentials/reducers/send_credential_request.rs
nanderstabel: https://github.com/impierce/identity-wallet/pull/157#discussion_r1526166365

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
Norlock: Yes I thought about it when I was walking to the store that the compiler wouldn't allow you to forget it ;). Yeah lets keep it like this :+1:

identity-wallet/src/state/connections/mod.rs
nanderstabel: The implementation in this PR is not the definitive solution for this problem of identifying Issuers/RelyingParties, but it is just slightly better than the previous solution:

"name" is not unique, any Issuer can come up with a name, or duplicate another servers name.

previous solution = connection ID based on "name"
current solution = connection ID based on "name" AND "url" (host url)

There are three scenario's:
1. Same issuer, but with subdivisions
Issuer 1: { name = "subdivision_A", url = "issuer_a.com" }
Issuer 1:  { name = "subdivision_B", url = "issuer_a.com" }
previous solution --> two different connections :white_check_mark: 
current solution --> two different connections :white_check_mark:  

2. Issuers are completely different
Issuer 1: { name = "Issuer 1", url = "issuer_a.com" }
Issuer 2:  { name = "Issuer 2", url = "issuer_b.com" }
previous solution --> two different connections :white_check_mark: 
current solution --> two different connections :white_check_mark:  

3. Issuer 2 is trying to impersonate Issuer 1?
Issuer 1: { name = "Issuer 1", url = "issuer_a.com" }
Issuer 2:  { name = "Issuer 1", url = "issuer_b.com" } <-- potentially a scammer?
previous solution --> cannot make a distinction between Issuer 1 and Issuer 2 :x: 
current solution --> two different connections :white_check_mark: 


So to sum up:
* name is not inherently a unique identifier
* the host of a url is

nanderstabel: And to add a bit more: in this current change we use the hostname of the url as an identifier. So:
"https://university.com/subdivision_A" 's hostname --> `university.com`
"https://university.com/subdivision_b" 's hostname --> also `university.com`

So if the university wants to issue credentials from different subdivisions, then it will need the "name" identifier to make the correct distinction.

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
Oran-Dan: Yes agreed

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
Oran-Dan: You mean that there would be no more Vec<\HistoryEvent> field in the Appstate Joris? that this would be become a substruct of sorts to the Connection struct?

identity-wallet/src/state/connections/reducers/handle_siopv2_authorization_request.rs
```diff
@@ -98,18 +76,16 @@ pub async fn handle_siopv2_authorization_request(mut state: AppState, _action: A
...
 ) -> anyhow::Result<(String, Option<String>, String)> {
-    let connection_url = siopv2_authorization_request
-        .body
-        .redirect_uri
-        .domain()
-        .ok_or(anyhow::anyhow!("unable to get domain from redirect_uri"))?
-        .to_string();
+    // Get the connection url from the redirect url host (or use the redirect url if it does not
+    // contain a host).
+    let redirect_url = siopv2_authorization_request.body.redirect_uri.clone();
```
Norlock: Also use redirect_uri here.

Norlock (dismissed):

identity-wallet/src/state/connections/mod.rs
Oran-Dan: Clear, and i think the set-up as you set it out is great.
The scenario i'm specifying is actually one in which scenario 1 will happen so often that one student will have about 6 new connections per semester and (based on my previous study) about 18 connections per year from his Uni.
I do think this is a plausible future as Modular Education was one of the big keypoints in the KWC1 meet-up/workshop I attended nearly 5 months ago. 
Perhaps another issue but how do we prepare for this?

nanderstabel: There's no clear cut way of uniquely identify issuers/relyingparties in a way that covers all edge cases so think it's best to wait and see what will happen in the real world regarding these kind of things. We simply don't know yet whether a scenario like the one you describe would happen. 

But, since we're on the topic I also want to mention this:
In SIOPv2/OID4VP there actually **is** a standardized way of uniquely identify clients (Relying Parties) namely by the use of their `client_id` (Client Identifier). There are multiple different methods for this: https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#name-verifier-metadata-managemen
One of them is using DIDs for the Client Identifiers, which is what we currently focus on. 
In OID4VCI there is a standardized way of uniquely identifying Issuers, namely always by use of a case sensitive URL: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-13.html#name-credential-issuer-identifie

So that's great, their Identifiers are standardized, however, they're not fully compatible with each other. So when a University is both A RelyingParty, as well as an Issuer, then they would potentially communicate two different Identifiers with us, e.g. a DID and a case sensitive URL. So we would treat them as two separate Connections. 

There are some methods to validate that a DID is linked to a hostname: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-13.html#section-13.4-3.1
But this is optional and doesn't cover all of the issuer/client ID mismatches.

So the way of identifying Connections is definitely not optimal, but it works with NGDIL and the current real world application that are out there (which is very limited).

Oran-Dan (approved):
Looks good, glad to see some history testing included now as well

identity-wallet/src/state/connections/mod.rs
Oran-Dan: oke very nice all this in depth info, I will copy this convo to a discussion and resolve it for this PR

Oran-Dan (changes requested):
I have an inconsistency in the app.
Build on my laptop ubuntu, testing with "Shenron" profile I get the following:
Everytime I add NGDIL credentials again it does not show the connected event again.
Everytime I add KWC1 credentials again it Does show the connected event again.

Btw, I also still get '[2024-03-15][18:55:52][ERROR][identity_wallet::command] Operation timed out' sometimes?

Screenshot from 2024-03-15 19-31-53
Screenshot from 2024-03-15 19-31-49
Screenshot from 2024-03-15 19-31-38
Screenshot from 2024-03-15 19-31-27
Screenshot from 2024-03-15 19-31-18

Oran-Dan (dismissed):

identity-wallet/src/state/connections/mod.rs
```diff
@@ -1,30 +1,182 @@
...
+    /// Modelled after the `std::collections::HashMap::get_mut` method.
+    fn get_mut(&mut self, url: &str, name: &str) -> Option<&mut Connection> {
+        self.0
+            .iter_mut()
+            .find(|connection| connection.url == url && connection.name == name)
+    }
+
+    /// Inserts a new connection into the list of connections if it does not already exist. If it does exist, updates
+    /// the last interaction time and returns a reference to the connection.
+    pub fn insert_or_update(&mut self, url: &str, name: &str) -> &Connection {
```
daniel-mader: Since we're first checking if it exists and _then_ inserting if it doesn't, I would rename to `update_or_insert`.

nanderstabel: good point :sweat_smile:

identity-wallet/src/state/connections/mod.rs
```diff
@@ -1,30 +1,182 @@
...
+            info!("Updating existing connection: {} {}", name, url);
+            self.get_mut(url, name).map(|connection| {
+                connection.update_last_interaction_time();
+                &*connection
+            })
+        } else {
+            info!("Inserting new connection: {} {}", name, url);
+            self.insert(Connection::new(name.to_string(), url.to_string()))
+        }
+        .expect("Failed to insert or update connection")
```
daniel-mader: Please update the error message to `Failed to update or insert connection`

daniel-mader (approved):

Norlock (approved):
LGTM :1st_place_medal:

Issue #72, opened 2023-11-06
Add persistent (credentials and connections) sorted preference option


Oran-Dan:
### Description

Beside the search and sorting functionalities defined in issue #60, it would also be nice to have an option to have your credentials and connections "persistently" sorted. So that whenever the user opens this tab it is displayed already sorted according to the preference. Add sort component based on UX Designs (currently without the grid & list view).

### Motivation

It adds to the user experience.

### Resources
![★ V2 Dashboard](https://github.com/impierce/identity-wallet/assets/43676124/96adf3ec-af93-45b5-b9dc-7107158d7a96)

### Requirements

Exact structure of course open for discussion. From the top of my head, I had the following in mind:

Back-end
1. Adding a preferences field to the app_state struct as I can imagine more preferences will be added later on (such as date notation (American/European). Locale (language) can be added as a subfield to this struct as well. It should have a logical default: credentials: date added new-old, connections: last connected new-old.
2. After checking the sorting preference, sort the creds/connects in place in the app_state struct to keep it simple.
3. Make sure that whenever creds/connects are added/deleted it automatically resorts.

Front-end
1. Add this options to the settings tab as not to overcrowd the creds/connects tabs ( opinion of a back-end developer ;) )

### Are you planning to contribute this in a PR?

Yes

Oran-Dan:
Currently only need to move the old sorting algorithm to this persistent sorting. 
The older non-persistent sorting is not included in the new UX designs.
Then it should be implemented in the front-end.

PR #176, opened 2024-03-28
feat: introduce credential sorting 

SoulaAndrikopoulos:
# Description of change
We've introduced the sorting functionality to the credentials on the homepage.
 
- #72

We've added unit tests and integration tests


identity-wallet/bindings/ConnectionSorting.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+export type ConnectionSorting = { "name_az": { reverse: boolean | null, } } | { "first_interacted_new_old": { reverse: boolean | null, } } | { "last_interacted_new_old": { reverse: boolean | null, } };
```
daniel-mader: Should be generated as `reverse?: boolean` instead of `reverse: boolean | null` (same as in `UpdateSortingPreference.ts`). It's not necessary to fix it across the _entire_ repo in this PR, but only for the newly introduced bindings.

identity-wallet/bindings/Preferences.ts
daniel-mader, line 1: Is it possible to move some bindings to a subfolder inside `/bindings` such as `/bindings/sorting` or something? I feel like it's becoming harder to find things in the bindings folder if they're all next to each other.

identity-wallet/src/state/mod.rs
```diff
@@ -138,69 +131,69 @@ impl AppState {
...
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use crate::state::profile_settings::Locale;
-    use crate::state::profile_settings::Profile;
-    use indoc::indoc;
-    use tests::profile_settings::AppTheme;
+// #[cfg(test)]
```
daniel-mader: We should either make the test run again or remove it entirely, but not comment it out without further notice or comment _why_ it was commented.

unime/src-tauri/tests/fixtures/states/five_connections_sort_unicode_random_alphabets_name_az.json
```diff
@@ -0,0 +1,62 @@
...
+    {
+      "name": "s\u00e9cr\u00e9ter",
+      "url": "https://goverment.com",
+      "logo_uri": null,
+      "verified": false,
+      "first_interacted": "2023-11-09T13:52:11.986023042+00:00",
+      "last_interacted": "2023-11-09T13:52:19.971552835+00:00"
+    },
+    {
+      "id": "א",
```
daniel-mader, line 46: Should we have _actual_ unicode characters in our source code or can we somehow encode them? (I wouldn't know what that symbol is or how to recreate it)

unime/src/lib/connections/sorting/SortingSheetButton.svelte
```diff
@@ -0,0 +1,53 @@
...
+  let sortOrder: string;
+  $: {
+    if (method === 'name_az') {
+      sortOrder = reversed ? $LL.SORT.ORDER.Z_A() : $LL.SORT.ORDER.A_Z();
+    } else if (method === 'issue_date_new_old') {
+      sortOrder = reversed ? $LL.SORT.ORDER.NEWEST() : $LL.SORT.ORDER.OLDEST();
+    } else if (method === 'added_date_new_old') {
+      sortOrder = reversed ? $LL.SORT.ORDER.NEWEST() : $LL.SORT.ORDER.OLDEST();
+    } else {
+      sortOrder = ''; // Default to empty string if label doesn't match
```
daniel-mader: In what case would we have a default (empty string)?

unime/src/lib/connections/sorting/SortingSheetButton.svelte
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  // Dispatch click event
```
daniel-mader: I don't think that comment is necessary since the code below explains itself.

unime/src/lib/connections/sorting/SortingSheetButton.svelte
```diff
@@ -0,0 +1,53 @@
...
+
+<button
+  on:click={handleClick}
+  class={`my-1 flex w-full gap-2 border p-[10px] text-slate-800 dark:text-grey ${
+    active ? 'rounded-lg border-grey bg-silver dark:border-slate-600 dark:bg-navy' : 'border-transparent'
+  }`}
+>
+  <svelte:component this={icon} />
+  <p class="text-sm font-medium">{label}</p>
+  <!-- Render label -->
```
daniel-mader: I don't think we need those comments "Render xyz".

unime/src/lib/connections/sorting/SortingSheet.svelte
```diff
@@ -0,0 +1,75 @@
...
+    }
+    dispatch({
+      type: '[Settings] Update Sorting Preference',
+      payload: { credential_sorting, reverse: reversed },
+    });
+  }
+</script>
+
+<!-- bottom drawer and sorting button-->
+<div class="sort_wrapper">
```
daniel-mader: Where is this class `sort_wrapper` used?

unime/src/lib/connections/sorting/SortingSheet.svelte
```diff
@@ -0,0 +1,75 @@
...
+  import Issued from '~icons/ph/calendar-check';
+  import Added from '~icons/ph/calendar-plus';
+  import Slider from '~icons/ph/sliders-horizontal';
+
+  let method: CredentialSortMethod = $state.profile_settings.sorting_preferences.credentials.sort_method;
+  let reversed: boolean = $state.profile_settings.sorting_preferences.credentials.reverse;
+
+  function updateSortingPreference(credential_sorting: CredentialSortMethod) {
+    if (method == credential_sorting) {
+      reversed = !reversed;
```
maiertech: Use strict comparison === by default. Use == only when you have a reason.

unime/src/lib/connections/sorting/SortingSheetButton.svelte
maiertech: Maybe something like: // Dispatch component event based on DOM event.

unime/src/lib/connections/sorting/SortingSheetButton.svelte
```diff
@@ -0,0 +1,53 @@
...
+    } else if (method === 'added_date_new_old') {
+      sortOrder = reversed ? $LL.SORT.ORDER.NEWEST() : $LL.SORT.ORDER.OLDEST();
+    } else {
+      sortOrder = ''; // Default to empty string if label doesn't match
+    }
+  }
+
+  // Dispatch click event
+  const dispatch = createEventDispatcher();
+  const handleClick = () => {
```
maiertech: I would dispatch inline on line 37 to make it more concise.

unime/src/lib/static/svg/ArrowsDown.svelte
```diff
@@ -0,0 +1,9 @@
+<svg width="16" height="17" viewBox="0 0 16 17" fill="none" xmlns="http://www.w3.org/2000/svg">
```
maiertech: This file is not used anywhere.

unime/src/lib/static/AlphabeticalOrder.svelte
```diff
@@ -0,0 +1,17 @@
+<svg width="19" height="19" viewBox="0 -1 22 22" xmlns="http://www.w3.org/2000/svg">
```
maiertech: Should this file be located in the SVG subfolder?

unime/src/lib/credentials/Favorites.svelte
```diff
@@ -13,9 +11,12 @@
 
   export let credentialType: 'all' | 'data' | 'badges' = 'all';
 
-  let favorite_credentials: DisplayCredential[] = $state.credentials.filter((c) => c.metadata.is_favorite);
+  // Make favorite_credentials reactive in case we sort favorites in the future, too.
```
maiertech: Comment needs to be updated. After we made filtering of favorites reactive, they are sorted as well, because backend sorts all credentials.

unime/src/lib/components/molecules/navigation/Tabs.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
```
maiertech, line 1: There is a warning on line one: Semicolon or block is expected (svelte). Any idea what triggers it?

unime/package.json
```diff
@@ -32,6 +32,7 @@
     "@iconify-json/emojione": "^1.1.7",
     "@iconify-json/heroicons": "^1.1.11",
     "@iconify-json/lucide": "^1.1.112",
+    "@iconify-json/mdi": "^1.1.64",
```
maiertech: I don't see this used anywhere (searching for ~icons/mdi).

maiertech (changes requested):
I reviewed frontend only.

identity-wallet/bindings/CredentialSorting.ts
nanderstabel, line 1: I believe this binding is not being used, so can it be removed? Same goes for `ConnectionSorting`.

identity-wallet/bindings/ConnectionSorting.ts
nanderstabel, line 1: I believe this binding is not being used, so can it be removed? Same goes for `CredentialSorting`.

identity-wallet/src/state/profile_settings/mod.rs
```diff
@@ -53,6 +55,43 @@ pub enum Locale {
...
+#[ts(export)]
+pub enum ConnectionSortMethod {
+    #[default]
+    #[serde(rename = "name_az")]
+    NameAZ,
+    #[serde(rename = "first_interacted_new_old")]
+    FirstInteractedNewOld,
+    #[serde(rename = "last_interacted_new_old")]
+    LastInteractedNewOld,
+}
```
nanderstabel: `[resolved]`
Minor suggestion, feel free to ignore.

with serde you can rename all enum variants all at once according to a preferred casing style. You can still overwrite the 'default' renaming for individual variants (because the snake_case version of `NameAZ` would be `name_a_z` instead of `name_az`):

```rust
#[derive(Clone, Serialize, Debug, Deserialize, TS, Default, PartialEq, EnumString)]
#[ts(export)]
#[serde(rename_all = "snake_case")]
pub enum CredentialSortMethod {
    #[default]
    #[serde(rename = "name_az")]
    NameAZ,
    // #[serde(rename = "issue_date_new_old")]
    IssueDateNewOld,
    // #[serde(rename = "added_date_new_old")]
    AddedDateNewOld,
}

#[derive(Clone, Serialize, Debug, Deserialize, TS, Default, PartialEq, EnumString)]
#[ts(export)]
#[serde(rename_all = "snake_case")]
pub enum ConnectionSortMethod {
    #[default]
    #[serde(rename = "name_az")]
    NameAZ,
    // #[serde(rename = "first_interacted_new_old")]
    FirstInteractedNewOld,
    // #[serde(rename = "last_interacted_new_old")]
    LastInteractedNewOld,
}

```

identity-wallet/bindings/Preferences.ts
nanderstabel: :heavy_plus_sign: 

You can do this by replacing `#[ts(export)]` with `#[ts(export, export_to = "bindings/sorting/ConnectionSortMethod.ts")]` (example for `ConnectionSortMethod`, but the same logic applies for the other sorting related bindings of course :) )

nanderstabel: Don't forget that if you change the paths of the bindings that the frontend also needs to be updated in order to find them

identity-wallet/src/state/actions.rs
```diff
@@ -117,6 +118,8 @@ mod bindings {
         UpdateCredentialMetadata { payload: UpdateCredentialMetadata },
         #[serde(rename = "[User Journey] Cancel")]
         CancelUserJourney,
+        #[serde(rename = "[Settings] Update Sorting Preference")]
```
nanderstabel: I think it's nice if we keep the formatting style equal for each action, so 
`"[Settings] Update sorting preference"` instead of
`"[Settings] Update Sorting Preference"`

identity-wallet/src/state/mod.rs
nanderstabel: I think it's safe to remove this test. It's basically a 'legacy' test from the time when we did not have the integration test setup that we have now. The integration test already prove that the serialization works so this one can go, the `indoc` dependency can also be removed from the Cargo.toml file probably.

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -0,0 +1,462 @@
...
+
+pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppState, AppError> {
+    let mut credentials: Vec<DisplayCredential> = state.credentials.clone();
+    let preferences: Preferences<CredentialSortMethod> = state.profile_settings.sorting_preferences.credentials.clone();
+
+    let display_names: Vec<String> = credentials
+        .iter()
+        .map(|credential| credential.display_name.clone())
+        .collect();
+    let sorted_display_names = sort(display_names, state.profile_settings.locale.clone());
```
nanderstabel, line 70: This sorting functionality will always be performed regardless of the sorting preference, even though `sorted_display_names` is only relevant for the `CredentialSortMethod::NameAZ` method. 

There's no need to optimize this now, but I think it would be good to add a `TODO` comment here same goes for the `sorted_names` in `fn sort_connections`.


(I know `name_az` is also used  in this block:
```rust
    if preferences.sort_method == CredentialSortMethod::IssueDateNewOld {
        let (mut credentials_with_date_issued, mut credentials_without_date_issued) = credentials
            .into_iter()
            // splits the `credentials` into two vectors.
            .partition::<Vec<_>, _>(|credential| !credential.metadata.date_issued.is_empty());

        credentials_without_date_issued.sort_by(name_az);
        credentials_with_date_issued.append(&mut credentials_without_date_issued);
        // assign the value reconstructed credentials back to `credentials`.
        credentials = credentials_with_date_issued
    };
```
But still, this can and should be optimized eventually).

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -0,0 +1,462 @@
...
+/// Sort a Vector of words alphabetically, taking into account the locale of the words:
+/// standard `.sorted()` -> ["Zambia", "enlever", "zoo", "énigme"]
+/// `sort(list, locale)` -> ["énigme", "enlever", "Zambia", "zoo"]
+/// Also making it Unicode compatible, using NFC-normalized form.
+pub fn sort(list: Vec<String>, locale: Locale) -> Vec<String> {
+    // https://github.com/unicode-org/icu4x/tree/main/components/collator#examples
+
+    // Convert locale from AppState to icu_locid::Locale
+    let locale: &str = locale.into();
+    let locale: icu::locid::Locale = locale.replace('_', "-").parse().unwrap();
```
nanderstabel: There's no need to replace here I believe. Both formats will be parsed equally:
```rust

#[test]
fn test() {
    let locale1: icu::locid::Locale = "en-GB".parse().unwrap();
    let locale2: icu::locid::Locale = "en_GB".parse().unwrap();

    assert_eq!(locale1, locale2);
}
```

nanderstabel: `[resolved]` 

Not relevant for now but perhaps a nice problem to work on on a blue Monday.. I feel like our own `Locale` enum could serve as a sort of wrapper around the `icu::locid::Locale` struct in a way that when we deserialize `"en-GB"` into our own `Locale` type, under the hood it utilizes `icu::locid::Locale`'s `FromStr` implementation. 

This way we can get rid off the `unwrap` here and we don't need the double conversion, first from `Locale` to `&str`, and then from `&str` to `icu::locid::Locale`.

unime/package.json
SoulaAndrikopoulos: I believe it was used for an icon, but not anymore. Will remove it 👍

unime/src/lib/components/molecules/navigation/Tabs.svelte
SoulaAndrikopoulos: Thanks for noticing this one! I am not sure at the moment, but will look into this.

unime/src/lib/connections/sorting/SortingSheet.svelte
SoulaAndrikopoulos: Removed it. ✅

unime/src/lib/connections/sorting/SortingSheetButton.svelte
SoulaAndrikopoulos: Removed it, it's not needed after all.

unime/src/lib/connections/sorting/SortingSheet.svelte
daniel-mader: You can remove the entire `div` if it's only wrapping one component (in this case `<ActionSheet />`) and doesn't have any classes.

unime/src/lib/components/molecules/navigation/Tabs.svelte
maiertech: We leave it as is for the time being.

maiertech (dismissed):
Frontent portion is ready to be merged.

identity-wallet/bindings/Preferences.ts
Oran-Dan: updated

identity-wallet/src/state/mod.rs
Oran-Dan: updated

unime/src-tauri/tests/fixtures/states/five_connections_sort_unicode_random_alphabets_name_az.json
Oran-Dan: we have actual unicode in the tests, as explained in the comments:
```
    // Example of unicode characters, and how they are understood differently by the computer.
    // This example is also used in the json files
    // let no_unicode = "sécréter";
    // let unicode = "sécréter";
    // assert_ne!(no_unicode, unicode);
    // assert_ne!(no_unicode.len(), unicode.len());
    // The above is directly written as unicode characters.
    // In unicode sequence it is "s\u{00e9}cr\u{00e9}ter"
    ```
Does this resolve your question?

File: identity-wallet/bindings/CredentialSorting.ts
Comment (Oran-Dan) [reply identity-wallet/bindings/CredentialSorting.ts:1]: updated

File: identity-wallet/bindings/ConnectionSorting.ts
Comment (Oran-Dan) [reply identity-wallet/bindings/ConnectionSorting.ts:1]: updated

File: identity-wallet/src/state/profile_settings/mod.rs (hunk 15 above)
Comment (Oran-Dan) [reply identity-wallet/src/state/profile_settings/mod.rs]: updated

Comment (Oran-Dan) [reply identity-wallet/src/state/profile_settings/mod.rs]: also for AppTheme

File: identity-wallet/src/state/actions.rs (hunk 16 above)
Comment (Oran-Dan) [reply identity-wallet/src/state/actions.rs]: updated

File: identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs (hunk 18 above)
Comment (Oran-Dan) [reply identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs]: updated

Comment (Oran-Dan) [reply identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs]: you are right, this would be cleaner, i'll keep it in an issue for refactoring the codebase

File: identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs (hunk 17 above)
Comment (Oran-Dan) [reply identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs:70]: Yes, you're right, I haven't found the time yet to dig into the error message when moving it into the relevant code block:
```
expected a closure that implements the `FnMut` trait, but this closure only implements `FnOnce`
this closure implements `FnOnce`, not `FnMut`
```
I've added the TODO and will look into this another time

nanderstabel (dismissed):
Great work! :fireworks:

maiertech (dismissed):
Finally! Good job everyone!

daniel-mader (approved):

maiertech (approved):

Issue #178, opened 2024-04-04
[Bug] Locale is reset to default after onboarding


daniel-mader:
### Description

When selecting a locale during onboarding, it is reset to default (`en_US`) when the profile is created.
This line might be responsible for that behavior:
https://github.com/impierce/identity-wallet/blob/007d8246d9f36ea421cb4b286177d001ad1c33d3/identity-wallet/src/state/profile_settings/reducers/create_new.rs#L49

### Steps to Reproduce the Bug

1. select other language than `en_US` on first page of onboarding
2. complete onboarding
3. language is reset to `en_US`

### Expected Behaviour

Chosen language is not reset

### Actual Behaviour

see above

### Errors

n/a

PR #180, opened 2024-04-07
fix: prevent locale from being reset after onboarding

Oran-Dan:
# Description of change
I have changed the locale being reset to default after onboarding in the ```create_new``` reducer

- #178 

I've added a test in the /tests/get_state.rs file

### DIFF
--- a/unime/src-tauri/tests/common/assert_state_update.rs
+++ b/unime/src-tauri/tests/common/assert_state_update.rs
@@ -50,7 +50,7 @@ pub async fn assert_state_update(
         // Assert that the state is updated as expected.
         if let Some(expected_state) = expected_state {
             let container = app.app_handle().state::<AppStateContainer>().inner();
-            let mut guard = container.0.lock().await;
+            let guard = container.0.lock().await;
 
             let AppState {
                 connections,
@@ -62,7 +62,9 @@ pub async fn assert_state_update(
                 history,
                 extensions,
                 ..
-            } = &mut *guard;
+            } = &mut guard.clone();
+
+            drop(guard);
 
             let AppState {
                 connections: expected_connections,
@@ -75,6 +77,8 @@ pub async fn assert_state_update(
                 extensions: expected_extensions,
                 ..
             } = expected_state;
+
+            println!("Current state:\n{:#?}\n\n-------------------------------------\n\nExpected state:\n{:#?}\n", container.0.lock().await, expected_state);
 
             assert_eq!(connections, expected_connections);
 
--- a/unime/src-tauri/tests/common/assert_state_update.rs
+++ b/unime/src-tauri/tests/common/assert_state_update.rs
@@ -78,7 +78,11 @@ pub async fn assert_state_update(
                 ..
             } = expected_state;
 
-            println!("Current state:\n{:#?}\n\n-------------------------------------\n\nExpected state:\n{:#?}\n", container.0.lock().await, expected_state);
+            println!(
+                "Current state:\n{:#?}\n\n-------------------------------------\n\nExpected state:\n{:#?}\n",
+                container.0.lock().await,
+                expected_state
+            );
 
             assert_eq!(connections, expected_connections);
 
--- a/unime/src-tauri/tests/tests/get_state.rs
+++ b/unime/src-tauri/tests/tests/get_state.rs
@@ -57,7 +57,7 @@ async fn test_locale_persistency_during_onboarding() {
         container,
         vec![
             // Get the initial state.
-            action // Create a new profile.
+            action, // Create a new profile.
         ],
         vec![
             // The profile was created, so the user is redirected to the profile page.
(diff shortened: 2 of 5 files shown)


unime/src-tauri/tests/common/assert_state_update.rs
```diff
@@ -76,6 +78,12 @@ pub async fn assert_state_update(
                 ..
             } = expected_state;
 
+            println!(
```
daniel-mader, line 81: If you want to log, you should use macro from the `log` crate (such as `log::info`) instead of a `println!()`, but I'm not sure if it's a good idea to log the entire current and expected state for each single test.

unime/src-tauri/tests/tests/get_state.rs
```diff
@@ -39,6 +39,34 @@ async fn test_get_state_create_new() {
     .await;
 }
 
+#[tokio::test]
+#[serial_test::serial]
+async fn test_locale_persistency_during_onboarding() {
+    setup_state_file();
+    setup_stronghold();
+
+    // Deserializing the AppStates and Actions from the accompanying json files.
```
daniel-mader, line 48: I don't think this comment adds more meaning to the code since it is a quite common test setup.

unime/src-tauri/tests/tests/get_state.rs
```diff
@@ -39,6 +39,34 @@ async fn test_get_state_create_new() {
...
+    let state1 = json_example::<AppState>("tests/fixtures/states/no_profile_nl-NL_redirect_welcome.json");
```
daniel-mader, line 49: Can you give it a more speaking name instead of `state1`, such as `state_initial` or `state_before`? (Same applies for `state2` which could have a name such as `state_expected` or `state_after`)?

unime/src-tauri/tests/tests/get_state.rs
```diff
@@ -39,6 +39,34 @@ async fn test_get_state_create_new() {
...
+    let state2 = json_example::<AppState>("tests/fixtures/states/nl-NL_redirect_me.json");
+    let action = json_example::<Action>("tests/fixtures/actions/create_new.json");
+
+    let container = AppStateContainer(Mutex::new(state1));
+
+    assert_state_update(
+        // Initial state.
+        container,
+        vec![
+            // Get the initial state.
```
daniel-mader, line 59: The information carried by those comments could be expressed by more meaningful variable names (see above).

unime/src-tauri/tests/tests/get_state.rs
```diff
@@ -39,6 +39,34 @@ async fn test_get_state_create_new() {
     .await;
 }
 
+#[tokio::test]
+#[serial_test::serial]
+async fn test_locale_persistency_during_onboarding() {
```
daniel-mader: The test name is a little bit misleading: "during onboarding"? I think it would be more fitting to give it a name such as `locale_stays_unchanged_on_profile_creation` because that's what the code actually does, right?

Oran-Dan: sure, changed it

unime/src-tauri/tests/tests/get_state.rs
Oran-Dan: this is the standard format we use for every one of our tests. I definitely take your point but i rather stay consistent or change all tests accordingly

unime/src-tauri/tests/common/assert_state_update.rs
Oran-Dan: So the cargo test environment works a bit different. Logging won't show up.
And I am confident this is very helpful. It shows when it passes and when it fails. 
So immediately you can check if your positive is not a false positive.
And a negative, the assertion fail is very little information at times as well. 
This gives the full overview and it is only displayed on single tests, running cargo test --all won't bombard you with all the outputs.

unime/src-tauri/tests/tests/get_state.rs
Oran-Dan: see comment below

unime/src-tauri/tests/tests/get_state.rs
Oran-Dan: see comment above

unime/src-tauri/tests/common/assert_state_update.rs
nanderstabel: Log messages will not show up in `#[test]` context unless you utilize some test crate like [tracing_test](https://crates.io/crates/tracing-test). But I don't think that is really necessary though, since a simple `println` does the job I'd say (so we don't need to add `tracing_test` to all our tests).

If it aids to developer experience I think it is a good addition here.

unime/src-tauri/tests/tests/get_state.rs
nanderstabel: All the 'integration tests' in the format we're using right now were meant to be temporary since they were supposed to be generated from mermaid diagrams. So naming actions and states didn't really matter since it would have been abstracted away anyway. 

Since continuation of this is postponed to the 'hardening phase', as well as the fact that there are some other considerations to be made about the current 'testing format' imo it is better to be consistent now, and leave all testing framework specific considerations until we've (time to) come up with something better.

nanderstabel (approved):

daniel-mader (approved):

PR #186, opened 2024-04-10
fix: use hashed `logo_uri` as identifier for temporary image assets

daniel-mader:
# Description of change
Using a static name for issuer logos (`client_0.svg`, `client_0.png`) might have caused unintended caching in the frontend when querying image assets from the Tauri core ("backend"). This was never an issue in local development (possibly due to inactive caching), but resulted in previous issuer images being displayed when others were expected. Hashing the path to the original image should make the temporary image names unique and thereby solve the caching issue.

n/a

Physical iOS device


nanderstabel (dismissed):
Looks good, I have no further comments :)

Oran-Dan (changes requested):
Nice work! just the helper fn should be moved, in my opinion.

identity-wallet/src/state/mod.rs
```diff
@@ -138,6 +138,10 @@ impl AppState {
     }
 }
 
+fn hash(url: &str) -> String {
```
Oran-Dan: This should be moved to either `identity-wallet/src/state/core_utils/helpers.rs` or `identity-wallet/src/persistence.rs`. I prefer the first since although this helper fn is mainly concerned with the assets and thos functions live in the persistence file, this helper is not actually used in that file but throught multiple reducers so it should be in core_utils/helpers i think.

Oran-Dan (approved):

nanderstabel (approved):

PR #191, opened 2024-04-15
feat(backend): add sorting based on locale

Oran-Dan:
# Description of change
I've added a helper function to sort names respective to the locale selected by the user.
The sorting is also Unicode (NFC normalization form) compatible.
I've added unit and integration testing for the new functionality.
I've implemented the sorting function in all actions which change the connections/credentials field.

Resources:
https://sts10.github.io/2023/01/29/sorting-words-alphabetically-rust.html
https://www.unicode.org/reports/tr15/#Norm_Forms

- #72 

Tests can be found in `identity-wallet/unime/src-tauri/tests/tests/sorting.rs`


identity-wallet/src/state/connections/actions/connection_accepted.rs
```diff
@@ -2,7 +2,8 @@ use crate::{
     reducer,
     state::{
         actions::ActionTrait,
-        connections::reducers::handle_siopv2_authorization_request::handle_siopv2_authorization_request, Reducer,
+        connections::reducers::handle_siopv2_authorization_request::handle_siopv2_authorization_request,
```
Norlock, line 5: Maybe we need to improve the exports, its kinda verbose :)

nanderstabel:
Looks like your branch has some conflicts that need to be resolved first.

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -58,45 +60,91 @@ pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppSta
...
-    let last_interacted_new_old = |a: &Connection, b: &Connection| a.last_interacted.cmp(&b.last_interacted);
+    let list: Vec<String> = connections.iter().map(|x| x.name.clone()).collect();
+    let sorted_list = sort(list, state.profile_settings.locale.clone());
 
-    connections.sort_by(match preferences.sort_method {
-        ConnectionSortMethod::NameAZ => name_az,
-        ConnectionSortMethod::FirstInteractedNewOld => first_interacted_new_old,
-        ConnectionSortMethod::LastInteractedNewOld => last_interacted_new_old,
-    });
+    let name_az = |a: &Connection, b: &Connection| {
```
Norlock, line 122: I see that this function is also defined in the function above maybe we can create one function and reuse.

unime/src-tauri/tests/common/assert_state_update.rs
```diff
@@ -62,7 +62,9 @@ pub async fn assert_state_update(
                 history,
                 extensions,
                 ..
-            } = &mut *guard;
+            } = &mut guard.clone();
```
Norlock: if cloned why is it borrow mutably? I think you can remove &

unime/src-tauri/tests/fixtures/states/credential_redirect_me.json
```diff
@@ -37,6 +37,11 @@
           "birthdate": "1985-05-21"
         }
       },
+      "metadata": {
+        "is_favorite": false,
+        "date_added": "",
+        "date_issued": "\"2022-01-01T00:00:00Z\""
```
Norlock: There shouldn't be double quotes in a date field in json

unime/src-tauri/tests/fixtures/states/credential_share_credential.json
```diff
@@ -27,6 +27,11 @@
           "birthdate": "1985-05-21"
         }
       },
+      "metadata": {
+        "is_favorite": false,
+        "date_added": "",
+        "date_issued": "\"2022-01-01T00:00:00Z\""
```
Norlock: Same here

unime/src-tauri/tests/fixtures/states/four_credentials_redirect_me.json
```diff
@@ -83,6 +62,11 @@
           "validity": "Valid"
         }
       },
+      "metadata": {
+        "is_favorite": false,
+        "date_added": "",
+        "date_issued": "\"2022-08-15T09:30:00Z\""
```
Norlock: same here

Norlock (changes requested):
Try to un-double all the date strings in the JSON, and you can ignore the TODO ;).

identity-wallet/src/state/connections/actions/connection_accepted.rs
Oran-Dan: yes true, how do you envision such an improvement?

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
Oran-Dan: you mean in the Credential function? because it is slightly different there since the enum the match case is based on is a different enum, so i don't really see a convenient way to abstract a general function here

unime/src-tauri/tests/fixtures/states/credential_redirect_me.json
Oran-Dan: fixed it

unime/src-tauri/tests/fixtures/states/four_credentials_redirect_me.json
Oran-Dan: fixed

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -95,7 +96,7 @@ impl From<CredentialFormats<WithCredential>> for VerifiableCredentialRecord {
                     metadata: CredentialMetadata {
                         is_favorite: false,
                         date_added: DateUtils::new_date_string(),
-                        date_issued: issuance_date.to_string(),
+                        date_issued: issuance_date.to_string().replace('\"', ""),
```
daniel-mader: If it's only about removing the encapsulating quotes, we should tackle the root of the problem and parse it to an actual `DateTime`. I think the problem stems from calling `.to_string()` on a `serde_json::Value` which wraps it in string quotes. Might be worth investigating and solve it cleanly rather than applying more bandaid here. 👍

Oran-Dan: yes that's the problem. Changing that field to a DateTime seems like a good option. But to be consistent we would have to do that to all our timestamp fields (added, issued, first interaction, last "). I'll make it into another issue if we agree.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -25,22 +27,41 @@ pub struct DisplayCredential {
...
+            format: serde_json::from_str("\"test\"").unwrap(),
+            data: Default::default(),
+            metadata: CredentialMetadata::default(),
+            display_name: Default::default(),
+        }
+    }
+}
+
+#[test]
+fn test() {
```
daniel-mader: This test doesn't assert anything. Can it be removed?

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -25,22 +27,41 @@ pub struct DisplayCredential {
...
 
 #[typetag::serde(name = "display_credential")]
 impl FeatTrait for DisplayCredential {}
 
+impl Default for DisplayCredential {
+    fn default() -> Self {
+        Self {
+            id: Default::default(),
+            issuer_name: Default::default(),
+            format: serde_json::from_str("\"test\"").unwrap(),
```
daniel-mader, line 39: Can you leave a comment why "test" is the default?

package-lock.json
daniel-mader, line 1: Was this file added on purpose?

identity-wallet/bindings/actions/UpdateSortingPreference.ts
daniel-mader, line 1: Where's the difference in the file other than an empty line added at the end of the file? Since this is generated, I don't think this change needs to be added.

unime/src/lib/connections/sorting/SortingSheet.svelte
```diff
@@ -2,6 +2,9 @@
   import type { CredentialSortMethod } from '@bindings/CredentialSortMethod';
   import { melt } from '@melt-ui/svelte';
 
+  //import Ascending from '~icons/ph/sort-ascending';
```
daniel-mader: This should not be added

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -126,3 +150,14 @@ fn get_type_name_from_data(credential_display: &serde_json::Value) -> Option<Str
         _ => None,
     }
 }
+
+////////////////////////////
+
+pub fn json_example<T>(path: &str) -> T
```
daniel-mader: Is this added on purpose?

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -6,13 +6,18 @@ use crate::{
...
         },
         AppState,
     },
 };
 
+use icu::collator::*;
 use log::debug;
+use unicode_normalization::UnicodeNormalization;
+
+// Frontend shouldn't persist reverse setting on any of the sorting options which aren't selected.
```
daniel-mader: This can be rewritten to "Reverse setting of non-selected sorting options are not persisted and use the default (false) when selected."

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -58,45 +64,398 @@ pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppSta
     let mut credentials: Vec<DisplayCredential> = state.credentials.clone();
     let preferences: Preferences<CredentialSortMethod> = state.profile_settings.sorting_preferences.credentials.clone();
 
-    let name_az = |a: &DisplayCredential, b: &DisplayCredential| a.display_name.cmp(&b.display_name);
-    let issuance_new_old =
-        |a: &DisplayCredential, b: &DisplayCredential| a.metadata.date_issued.cmp(&b.metadata.date_issued);
-    let added_new_old =
-        |a: &DisplayCredential, b: &DisplayCredential| a.metadata.date_added.cmp(&b.metadata.date_added);
+    let list: Vec<String> = credentials.iter().map(|x| x.display_name.clone()).collect();
```
daniel-mader: I would pick more speaking names than the generic `list` and `x`.

daniel-mader: Such as `unsorted` and `credential` (or `c` if you wanna be short).

daniel-mader (changes requested):

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -81,7 +94,10 @@ impl From<CredentialFormats<WithCredential>> for VerifiableCredentialRecord {
                     )
                 };
 
-                let issuance_date = credential_display["issuanceDate"].clone();
+                let mut issuance_date = String::new();
+                if !credential_display["issuanceDate"].is_null() {
+                    issuance_date = credential_display["issuanceDate"].clone().to_string().replace('\"', "");
+                }
```
nanderstabel: This looks a bit over-complicated. The same functionality can be achieved using:
```rust
    let issuance_date = credential_display["issuanceDate"].as_str().unwrap_or_default();
```

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -58,45 +63,401 @@ pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppSta
...
-    credentials.sort_by(match preferences.sort_method {
-        CredentialSortMethod::NameAZ => name_az,
-        CredentialSortMethod::IssueDateNewOld => issuance_new_old,
-        CredentialSortMethod::AddedDateNewOld => added_new_old,
-    });
+    let name_az = |a: &DisplayCredential, b: &DisplayCredential| {
+        let cmp_a = sorted_display_names
+            .iter()
+            .position(|display_name| display_name == &a.display_name)
+            .unwrap();
```
nanderstabel: The `unwrap` here and in line 80 can be removed. `Option<usize>` can be compared as well and removing those unwraps removes any chance there might be of a panic.

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -58,45 +63,401 @@ pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppSta
...
         credentials.reverse();
     }
 
+    // In this block we check if there are credentials without an issue-date.
+    // When the issue-date is empty, we sort these credentials alphabetically and add them to the bottom.
+    if preferences.sort_method == CredentialSortMethod::IssueDateNewOld {
+        let mut credentials_empty_issue_date: Vec<DisplayCredential> = credentials
+            .iter()
+            .filter(|credential| credential.metadata.date_issued.is_empty())
+            .cloned()
```
nanderstabel: This could potentially be quite an expensive operation because it will clone all `DisplayCredential`s without a `date_issued`. This data type can be rather large especially because of its `data` field. 

I would suggest to came up with a solution here that doesn't clone any `DisplayCredential` e.g.:
```rust
    if preferences.sort_method == CredentialSortMethod::IssueDateNewOld {
        let (mut credentials_with_date_issued, mut credentials_without_date_issued) = credentials
            .into_iter()
            // splits the `credentials` into to vectors.
            .partition::<Vec<_>, _>(|credential| !credential.metadata.date_issued.is_empty());

        credentials_without_date_issued.sort_by(name_az);
        credentials_with_date_issued.append(&mut credentials_without_date_issued);
       // assign the value reconstructed credentials back to `credentials`.
        credentials = credentials_with_date_issued
    };
```

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -58,45 +63,401 @@ pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppSta
...
 
-    let name_az = |a: &Connection, b: &Connection| a.name.cmp(&b.name);
-    let first_interacted_new_old = |a: &Connection, b: &Connection| a.first_interacted.cmp(&b.first_interacted);
-    let last_interacted_new_old = |a: &Connection, b: &Connection| a.last_interacted.cmp(&b.last_interacted);
+    let names: Vec<String> = connections.iter().map(|connection| connection.name.clone()).collect();
+    let sorted_names = sort(names, state.profile_settings.locale.clone());
+
+    let name_az = |a: &Connection, b: &Connection| {
+        let cmp_a = sorted_names.iter().position(|name| name == &a.name).unwrap();
+        let cmp_b = sorted_names.iter().position(|name| name == &b.name).unwrap();
```
nanderstabel: The `unwrap`s can be omitted here as well.

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -58,45 +63,401 @@ pub async fn sort_credentials(state: AppState, _action: Action) -> Result<AppSta
...
-    let name_az = |a: &DisplayCredential, b: &DisplayCredential| a.display_name.cmp(&b.display_name);
-    let issuance_new_old =
-        |a: &DisplayCredential, b: &DisplayCredential| a.metadata.date_issued.cmp(&b.metadata.date_issued);
-    let added_new_old =
-        |a: &DisplayCredential, b: &DisplayCredential| a.metadata.date_added.cmp(&b.metadata.date_added);
+    let display_names: Vec<String> = credentials
+        .iter()
+        .map(|credential| credential.display_name.clone())
+        .collect();
+    let sorted_display_names = sort(display_names, state.profile_settings.locale.clone());
```
nanderstabel, line 70: `[resolved]`
This functionality is only relevant for the `CredentialSortMethod::NameAZ` preference, but it will also always be executed when one of the other sorting method preferences is set which is a bit inefficient. 

I think a future improvement would be to somehow solve that.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -25,22 +24,36 @@ pub struct DisplayCredential {
...
 
 #[typetag::serde(name = "display_credential")]
 impl FeatTrait for DisplayCredential {}
 
+impl Default for DisplayCredential {
+    fn default() -> Self {
+        Self {
+            id: Default::default(),
+            issuer_name: Default::default(),
+            // Here we use "test" as a placeholder since this fied has no default and it cannot be left empty.
```
daniel-mader: Typo: "fied" --> "field"

identity-wallet/src/state/profile_settings/reducers/update_sorting_preference.rs
```diff
@@ -6,13 +6,17 @@ use crate::{
...
         },
         AppState,
     },
 };
 
+use icu::collator::*;
 use log::debug;
+use unicode_normalization::UnicodeNormalization;
+
+//Reverse setting of non-selected sorting options are not persisted and use the default (false) when selected.
```
daniel-mader: Missing space at the start of the comment: `//Reverse`

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -94,10 +94,11 @@ impl From<CredentialFormats<WithCredential>> for VerifiableCredentialRecord {
...
 
-                let mut issuance_date = String::new();
-                if !credential_display["issuanceDate"].is_null() {
-                    issuance_date = credential_display["issuanceDate"].clone().to_string().replace('\"', "");
-                }
+                let issuance_date = credential_display["issuanceDate"]
+                    .as_str()
+                    .map(ToString::to_string)
+                    .unwrap_or_default()
+                    .replace('\"', "");
```
nanderstabel: TL;DR: don't need `.replace('\"', "")` here.

this is not necessary, there won't be an instance where you'll encounter `'\"'`.

In your previous solution you needed it because you made use of the `serde_json::Value::to_string()` function which may result in somewhat confusing behavior. It is `serde_json::Value::to_string()` that produces those `'\"'`, so you don't need to worry about them if you don't use that method.

```rust
#[test]
fn test() {
    let credential_display = json!({
        "issuanceDate": "2022-00-00T00:00:00Z"
    });

    // this will result in: `"\"2022-00-00T00:00:00Z\""`
    let a = credential_display["issuanceDate"].to_string(); // <-- a bandaid solution here is to add `.replace('\"', "")` at the end

    // this will result in `"2022-00-00T00:00:00Z"`
    let b = credential_display["issuanceDate"]
        .as_str()
        .map(ToString::to_string)
        .unwrap();

    // So this assertion will fail, because they're not the same.
    assert_eq!(a, b);
}
```

nanderstabel (approved):

daniel-mader (approved):

identity-wallet/src/state/credentials/mod.rs
daniel-mader: https://github.com/impierce/identity-wallet/issues/204

PR #220, opened 2024-04-30
fix: Change color of inactive bottom nav items

### LINKED ISSUE #219: [Bug] BottomNavItems look disabled
### Description

This issue fixes the display color of inactive items in the `BottomNavBar`. This issue is a spin-off from https://github.com/impierce/identity-wallet/issues/211.

### Hardware Specification

n/a

Navigate to the credentials overview `/me`.

### Expected Behaviour

Inactive buttons should be rendered in a way that they do not appear to be disabled.

### Actual Behaviour

The `BottomNavBar` contains 2 `BottomNavItem`s. The active one is highlighted with a green bar, but the two inactive ones are rendered with a pale gray. This makes them look like they are disabled:

App users might assume that inactive items cannot be tapped.

### Errors

n/a

maiertech:
# Description of change

- This PR changes the color of inactive bottom navigation items. Refer to #219.
- I cleaned up SVG files and use `fill="currentColor"` for styling.
- I removed conditional styles, since the color is now the same for active and inactive bottom navigation items.

Fixes #219.

Manual visual test on the credentials page on the `/me` route.


Oran-Dan (commented):
I like it, but I do wonder if it might be considered inconsistent with the tabs. The unselected tabs are still greyed out, do we want those white then as well? I'm biased since i'm used to the old style so i'm not sure if new users would even notice this difference between unselected tabs on the bottom or the "All, Data, Badges" tabs, so just wondering. Curious to here what you guys think about it?

Also wondering, as I take a second look, if just lightly less white for the unselected tabs would be a bit less busy on the screen than all the same white.

SoulaAndrikopoulos:
I agree with the points that Oran brought up. 👍 It looks nice, but we have to consider consistency with the tabs (so we follow either of the patterns; greyed out or white). If we give the tabs a white colour it would look as follows:

Screenshot 2024-05-02 at 12 27 43

At the moment, I don't have a clear preference between this and the older design, but I'd be curious to hear other's opinions. :)

maiertech:
The question is whether a user perceives the bottom navigation as tabs. We should run this by Andrew.

daniel-mader:
I took a look at how other popular apps style their nav bars and this is what I got for GitHub, Mastodon & Slack (all iOS). I would sum this up with: active icon is in primary color, others are _slightly_ greyed out, no indicator bar. GitHub Mastodon Slack

daniel-mader:
If we apply the same styling to the UniMe nav bar, it would look as follows: 
light-green-none-cut
dark-green-none-cut
I would personally prefer this over the currently proposed update (this PR) which is only relying on the indicator:
Screenshot 2024-05-07 at 10 53 03
Screenshot 2024-05-07 at 10 53 21

unime/src/lib/components/molecules/navigation/BottomNavItem.svelte
```diff
@@ -13,16 +13,12 @@
...
-      <!-- Currently, there is no better way of applying conditional Tailwind classes to a named slot inside a component. -->
+    <div class="flex h-12 w-10 flex-col items-center justify-center text-slate-800 dark:text-grey">
+      <!-- Slot accepts SVG icons, which bring their own dimensions. -->
       <slot name="icon" />
-      <div
-        class={`pt-[6px] text-[12px]/[20px] font-medium ${
-          active ? 'text-slate-800 dark:text-grey' : 'text-slate-300 dark:text-slate-500'
-        }`}
-      >
+      <div class="pt-[6px] text-[12px]/[20px] font-medium">
```
daniel-mader, line 25: Should we really drop the styling of the icon based on "active" state? (_see screenshots above_)

daniel-mader (changes requested):

unime/src/lib/components/molecules/navigation/BottomNavItem.svelte
maiertech: Also fine. I would, however, choose a color with AAA contrast for inactive icons.

daniel-mader:
After a little consultation with @digital-zen, this would be the preferred outcome:

light-green-thin-2-cut

* reduce indicator thickness from 4px to 2px
* inactive items are colored in slate-800
* active item is colored in primary color
* dark theme is respected accordingly

daniel-mader (approved):

SoulaAndrikopoulos (approved):
The navbar icons are more noticeable, everything works well in the app, code looks more concise, and comments have been removed. 👍

Issue #182, opened 2024-04-08
[Bug] Image shows fallback slot while asset is loading


daniel-mader:
### Description

While an image is loaded from disk using the `<Image />` component, for a brief moment the fallback slot is rendered. While this works as designed, it can confuse the user since the image is _almost_ instantly loaded (aka "placeholder flashing").

### Steps to Reproduce the Bug

1. navigate around the app and pay attention to wherever images are displayed

### Expected Behaviour

Images should be preloaded before the navigation is executed to prevent flashing placeholders, but this might introduce an artificial delay and introduce bad UX

### Actual Behaviour

https://github.com/impierce/identity-wallet/assets/45594838/6f6edb1c-b093-4a0b-a284-ed15d283b2d0

### Errors

n/a

maiertech:
Obsolete with `/credential` route redesign in #366.

PR #233, opened 2024-05-17
fix: reduce flicker when loading credential images

maiertech:
# Description of change

The previous handling of images loaded the fallback images immediately. The revised approach awaits a promise that returns the asset URL. The fallback image is shown only when no asset URL could be resolved.

If the asset URL can be resolved, the image is still rendered with a slight delay.

I tried an alternative approach to compute the asset URL in a SvelteKit page load function. But this did not work. Not clear why, but calling Tauri APIs crashes the load function.

The perceived flickering should be less because the fallback does not get rendered by default. But it is possible that you will perceive the delay when the image loads. This is unfortunately a limitation of client-side loading.

ScreenFlow

Fixes issue #182.

Screen record the app while clicking on badges in the Ferris profile. Then use the scrubber to scrub through the recorded images after clicking a badge. You should not see the fallback image loaded. However you will see no image loaded before that image gets loaded.

### DIFF
--- a/unime/src/lib/components/atoms/Image.svelte
+++ b/unime/src/lib/components/atoms/Image.svelte
@@ -21,16 +21,11 @@
   export let id: string;
   export let iconFallback: Icon = 'User';
   export let isTempAsset = false;
-  let assetUrl: string | null = null;
 
-  async function loadImage() {
-    getImageAsset(id, isTempAsset).then((url) => {
-      assetUrl = url;
-    });
-  }
+  let assetUrl: Promise<string | null>;
 
   onMount(() => {
-    loadImage();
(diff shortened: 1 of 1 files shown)


Norlock (approved):

maiertech:
I am switching this one back to draft because the fix needs to be tested on a real device first. Subject to getting `tauri ios dev` up and running on my machine.

maiertech:
The background image effect seems to be the culprit for the perceived flicker.

maiertech:
On hold until https://github.com/impierce/identity-wallet/issues/284 is done.

maiertech:
Obsolete with `/credential` route redesign in #366.

PR #325, opened 2024-08-07
fix: update logic to determine credential display name

### LINKED ISSUE #324: Credential name and issuer name are not correctly fetched from Credential Iissuer Metadata
### Description

When scanning a Credential Offer the Credential name and Issuer name are correctly read and displayed as can be seen in the first screenshot.

Screenshot from 2024-08-07 15-45-53

However, after accepting the credential, these names are not displayed anymore as can be seen below:

Screenshot from 2024-08-07 15-46-03

Somehow, both values are not properly fetched and and persisted (probably somewhere in `send_credential_request.rs`.

### Motivation

These values are very important display properties that not to be properly displayed

### Resources

n/a

### To-do List

- [ ] Fix Credential Issuer name
- [ ] Fix Credential name

nanderstabel:
# Description of change
For more details, see #324 

This PR also fixes some flaky integration tests. Some of the tests may throw an error sometimes:
```sh
---- tests::get_state::test_locale_stays_unchanged_on_profile_creation stdout ----
thread 'tokio-runtime-worker' panicked at identity-wallet/src/stronghold.rs:131:14:
stronghold could not commit: SnapshotFileMissing("Parent directory of snapshot file does not exist")
```

This has been resolved by adding (re-adding) the following to the flaky tests:
```rust
    setup_stronghold();
```

fixes #324 

Added the following unit tests:
* `display_name_is_successfully_read_from_credential_configuration`
* `display_name_is_successfully_read_from_credential_type`
* `display_name_is_successfully_read_from_credential_type_array`

Also, tested it manually against UniCore.


identity-wallet/src/state/credentials/reducers/send_credential_request.rs
```diff
@@ -271,3 +282,144 @@ pub async fn send_credential_request(state: AppState, action: Action) -> Result<
...
+                .or(
+                    // Or, if the `type` property is an array, use the last element as the name of the credential.
+                    verifiable_credential_record.display_credential.data["type"]
+                        .as_array()
+                        .and_then(|types| types.last())
+                        .and_then(|last_type| last_type.as_str()),
+                )
+        })
+        .map(ToString::to_string)
+        // An empty string is used as a fallback. This can only happen if the credential is not a valid W3C
```
daniel-mader: I think we should not fallback to an empty string as it might be confusing in the frontend (even if it's unlikely). Let's use some value such as "Unknown" or "Credential".

daniel-mader (commented):
I would suggest a renaming of the PR to something like `fix: update logic to determine credential display name` (since we're not actually "fetching" anything)

identity-wallet/src/state/credentials/reducers/send_credential_request.rs
nanderstabel: Fixed :)

daniel-mader (approved):

Oran-Dan (approved):

Issue #284, opened 2024-07-15
Consolidate `/badges` and `/credentials`routes


maiertech:
### Description

Both routes essentially render the same thing. `/badges` is the newer design. The consolidated route should be called `/credentials`.

### Resources

The refactored route should also fix https://github.com/impierce/identity-wallet/pull/233.

### To-do List

n/a

PR #351, opened 2024-08-30
feat: render embedded images in credential subject

### LINKED ISSUE #349: Add support for credential embedded images
### Description

UniMe currently does not support images embedded inside Credentials as they will currently just be displayed as a long string value. 

We need to be able to render such embedded images in a nice way. 

In a next iteration we can improve the design further by implementing claim-specific display propterties: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.1.1.2-3.1.2.2.1. (See #350)

Media types to be supported:
- `image/jpeg`
- `image/png`
- `image/svg`
- Any other media types listed here: https://www.iana.org/assignments/media-types/media-types.xhtml#image

### Motivation

Better user experience

### Resources

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.1.1.2-3.1.2.2.1
https://www.iana.org/assignments/media-types/media-types.xhtml#image
- #350

### To-do List

- [ ] Display images

daniel-mader:
# Description of change
Render all `credentialSubject` claims that start with `data:image/` as an `` instead of displaying them as regular text fields.

- closes #349 

Visual inspection in Desktop mode.


daniel-mader:
Blocked by #284

unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+  export let key: string;
+  export let dataUrl: string;
+</script>
+
+<!--
+@component
+This renderer for `credentialSubject` fields renders an embedded image for a given [Data URL](https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data).
+-->
+<div class="flex flex-col items-center rounded-xl bg-background-alt px-4 py-3 text-[13px]/[24px]">
```
Comment (maiertech) [inline unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte]: Maybe rather not use `text-[13px]/[24px]` and switch to a default Tailwind text size? Would need to be done on the entire `/credentials` route.

unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte
```diff
@@ -0,0 +1,13 @@
...
+  <h2 class="w-full font-medium text-text-alt">{key}</h2>
+  <img src={dataUrl} alt="embedded" class="h-auto w-32 rounded-lg" />
```
Comment (maiertech) [inline unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte]: Image should be left aligned because everything else is left aligned. Also are we sure images will always have a 1:1 ratio? Constraining the height makes sense but it should probably be accompanied by `object-contain` like in the credential logo.

And why do we need `h-auto` when width is constrained?

unime/src/routes/credentials/[id]/CredentialHeader.svelte
```diff
@@ -19,12 +19,12 @@
 
 <!-- Stretch over parent horizontal padding with negative margins. -->
 <div class="relative -mx-4 flex flex-col items-center gap-4 bg-background-alt py-5">
-  <div class="grid h-40 w-40 place-items-center rounded-xl bg-background">
+  <div class="grid h-40 w-40 place-items-center rounded-xl bg-white">
```
Comment (maiertech) [inline unime/src/routes/credentials/[id]/CredentialHeader.svelte:23]: Add comment why we deviate from color scheme.

maiertech (changes requested):
Should the image renderer be more robust to images with different aspect ratios?

unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte
Comment (daniel-mader) [reply unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte]: True, but I think redefining typography across the entire app should be a separate issue. We can also think about defining a custom class for `text-[13px]/[24px]` since it has 67 occurrences in the entire project.

unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte
Comment (daniel-mader) [reply unime/src/routes/credentials/[id]/DataUrlImageRenderer.svelte]: I thought it would look better if the image is centered, since it visually aligns better with the credential image in the header. I now left aligned it, until we have a clearer idea of how and where we want to display embedded images.

maiertech (approved):
Should the image renderer be more robust to images with different aspect ratios?

nanderstabel (approved):

Issue #320, opened 2024-08-06
Upgrade Tauri to v2


maiertech:
### Description

Upgrade all Tauri dependencies to v2.

### Motivation

Tauri v2 will be released end of August.

### Resources

Info about how to migrate from beta to RC: https://v2.tauri.app/blog/tauri-2-0-0-release-candidate/.

### To-do List

The migration path is upgrade to latest beta and then run migration script when upgrading to RC. It's probably worth skipping RC releases and wait for v2 to be released.

M-Adam-Hus:
![](https://github.trello.services/images/mini-trello-icon.png) [Upgrade Tauri to v2](https://trello.com/c/eIJSpZWS/51-upgrade-tauri-to-v2)

PR #348, opened 2024-08-23
build: Replace `npm` with `pnpm`

maiertech:
# Description of change

Replace `npm` with `pnpm`. Update configs and generate new lock file.

## Links to any relevant issues

- Fixes #347.

## How the change has been tested

- CI tasks should not fail.
- UniMe should start up locally.

## Definition of Done checklist

Add an `x` to the boxes that are relevant to your changes.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #364, opened 2024-09-03
build(deps): upgrade Tauri from `beta` to stable

### LINKED ISSUE #224: [Bug] Clippy fails after update to Rust v1.78.0
### Description

The Rust lint tasks in the GitHub Action started failing today. This seems to be an issue introduced with [Rust v1.78.0](https://releases.rs/docs/1.78.0/).

### Hardware Specification

n/a

1. Run `rustup update`.
2. Verify that you have v1.78.0.
3. Run Clippy in `identity-wallet` and `unime/src-tauri`: `cargo clippy --all-targets --all-features -- -D warnings`.

Note: if you run `cargo clippy` you will see warnings only.

### Expected Behaviour

Clippy should run without throwing errors or warnings.

### Actual Behaviour

Clippy fails with errors and/or warnings.

### Errors

clippy

### LINKED ISSUE #283: Upgrade Tauri
### Description

Upgrade Tauri dependencies, both Rust and NPM.

NPM:
- `@tauri-apps/plugin-log`
- `@tauri-apps/api`

### Motivation

_No response_

### Resources

_No response_

### To-do List

n/a

daniel-mader:
# Description of change

Tauri v2 is now stable and we can safely upgrade from `beta`.

The official migration guide was followed: https://v2.tauri.app/start/migrate/from-tauri-2-beta/

- closes #283, #320
- closes #224
  - reverts https://github.com/impierce/identity-wallet/pull/225
- blocked until https://github.com/tauri-apps/tauri/pull/12290 is released

n/a


M-Adam-Hus:
[build(deps): upgrade Tauri from beta to stable](https://trello.com/c/COpogoJz/75-builddeps-upgrade-tauri-from-beta-to-stable)

maiertech:
This branch contains a regression in dev mode (dashed lines should be off):

regression

maiertech:
I observed https://github.com/impierce/identity-wallet/issues/481 during testing.

identity-wallet/src/state/actions.rs
```diff
@@ -21,6 +21,8 @@ pub fn listen<T: ActionTrait + Clone>(action: Action) -> Option<T> {
     action.downcast_arc::<T>().ok().map(|action| (*action).clone())
 }
 
+// TODO: remove this once we have a better way to export the TS types.
+
```
nanderstabel: This placement doesn't really make any sense anymore :sweat_smile: . This TODO message refers to the `mod bindings` block and has nothing to do with the doc comments that are now beneath it. 

I understand that you did this to silence a Clippy warning but I don't think this is the proper solution. 

I believe @Oran-Dan's intention with the code example to demonstrate how different parts of the Redux pattern come together in code (since these parts are (intentionally) separated in our codebase. They're doc comments because this simply adds nice color coding instead of having all code as plain comments.

I think there are (at least) 3 options here (instead of you solution):
1. Change the example from doc comments to 'normal' comments (we will loose the nice color coding though)
2. Move this code example elsewhere, for example to the `mod.rs` file in the same folder some place where it will not cause the same clippy warning (or do this in combination with option 3).
3. Add `#[allow(clippy::empty_line_after_doc_comments)]` above the code example to silence the clippy warning (since even though we care about Clippy Lints, we can still assess where we think its suggestions are suitable and where not. We don't even make use of `cargo docs` at all so I think this is a valid option as well).

Option 2 and 3 can both be used actually.

nanderstabel (changes requested):
I've thoroughly tested in UniMe and it seems to me that this upgrade fully works. 

The only difference I noticed is that running `cargo tauri dev` in the root of the repository does not work anymore for me. I need to run in it in the `/unime` folder in order for it to work. I'm pretty sure that's due to a change in the Tauri CLI so not relevant to this PR (also, it's not really a hindrance)

identity-wallet/src/state/actions.rs
daniel-mader: Option 3 was chosen since moving code examples around is not in scope of this PR

daniel-mader:
@nanderstabel 

> I've thoroughly tested in UniMe and it seems to me that this upgrade fully works.
> 
> The only difference I noticed is that running `cargo tauri dev` in the root of the repository does not work anymore for me. I need to run in it in the `/unime` folder in order for it to work. I'm pretty sure that's due to a change in the Tauri CLI so not relevant to this PR (also, it's not really a hindrance)

I just checked out the current `dev` branch and tried running `cargo tauri dev` in the root as well and it also doesn't work there for me. This could have already happened when we switched from `npm` to `pnpm` in https://github.com/impierce/identity-wallet/pull/348.

Oran-Dan (approved):
Everything works smooth, played around quite a bit and also completed the demo flow. ```cargo tauri dev``` from any folder hasn't worked for me since I reinstalled Linux on my laptop. I'm missing some Tauri CLI download I think. But it was never a problem for me since ```pnpm tauri dev``` in the root folder worked for me right from the start, no hassle at all. (I had a lot more hassle first time I installed linux on my then new laptop)

nanderstabel (approved):
I've tested again extensively (Android) and to me it looks like everything is working as expected. Great upgrade! :partying_face:

PR #387, opened 2024-10-18
feat: receive, display and share SD-JWT credentials

### LINKED ISSUE #523: Bug: Non-root trailing slash in `credential_issuer` URL breaks Credential Offer QR scan
### Description

When scanning a Credential Offer QR code with the UniMe Identity Wallet, the offer fails to process if the `credential_issuer` URI includes a non-root path ending with a trailing slash. For example, a URI such as `http://localhost:3033/extra/path/` causes the error, whereas a root URI like `http://localhost:3033/` does not.

**Steps to Reproduce:**
1. Configure a Credential Offer with a `credential_issuer` URI that includes an additional path (e.g. `http://localhost:3033/extra/path/`).
2. Generate and display the corresponding QR code.
3. Scan the QR code using the UniMe Identity Wallet.
4. Observe that the Credential Offer is not processed correctly.

https://github.com/user-attachments/assets/b87dc943-1d31-429c-ae88-3dd20e37d54d

**Expected Behavior:**
The UniMe Identity Wallet should process the Credential Offer QR code regardless of whether the `credential_issuer` URI includes extra path segments, as long as the URI is valid.

### Motivation

_No response_

### Resources

_No response_

### To-do List

- [ ] Ensure that Credential Offers is correctly processed

nanderstabel:
# Description of change
This PR adds support for SD-JWT credentials.

**Limitations**
* Key Binding is **not** supported yet
* Concealing (selective disclosure) is also **not** supported yet
* The backend 'disguises' SD-JWT (Display) credentials as W3C Verifiable Credentials so that the Frontend knows what to display.

**Depends on**
- https://github.com/impierce/openid4vc/pull/82

closes #523 (implicitly through https://github.com/impierce/openid4vc/pull/91)

Added unit tests:
* `test_verifiable_credential_record_try_from_jwt_vc_json`
* `test_verifiable_credential_record_try_from_sd_jwt_vc`

This feature can also be tested by:
1. Obtaining a SD-JWT PID credential from a PID Issuer
2. Login to https://deonlinedrogist.co.nl/ (using simple SIOPv2 login)
3. Select an item and click "Bestel Nu"
4. Scan the QR Code containing the SD-JWT Presentation Request
5. Confirm sharing the Credential


M-Adam-Hus:
[Feat/sd jwt](https://trello.com/c/vgLjnudI/74-feat-sd-jwt)

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -54,45 +59,86 @@ impl TryFrom<serde_json::Value> for VerifiableCredentialRecord {
...
+                let hash = sha256::digest(json!(credential_subject).to_string());
+
+                let id = Uuid::from_slice(&hash.as_bytes()[..16])?.to_string();
+
+                let format = CredentialFormats::VcSdJwt(());
+                // TODO: Remove this workaround that is basically a way of disguising the SD JWT VC as a VC so that
+                // it can be displayed in the Frontend.
+                let data = json!({
+                    "type": ["VerifiableCredential"],
+                    "issuer": sd_jwt_vc.claims().iss,
```
Oran-Dan, line 92: don't we potentially lose quite some info here as well? there are more claims than only `iss`

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -54,45 +59,86 @@ impl TryFrom<serde_json::Value> for VerifiableCredentialRecord {
...
+                let credential_display = get_unverified_jwt_claims(&verifiable_credential)?["vc"].clone();
+
+                // TODO: We are using this hash as Credential ID so that we can prevent credential duplication in
+                // demo situations. Now we can actually delete Credentials in UniMe we don't need to use the hash of the
+                // credential as the ID anymore. We should simply generate a random UUID.
+                // Derive the hash from the credential display.
+                let hash = {
+                    let type_value = credential_display["type"].clone();
+
+                    let credential_subject_key = "credentialSubject";
```
Oran-Dan: You removed the `type_key` but kept the `credential_subject_key`, this seems inconsistent I think

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -105,3 +151,108 @@ impl TryFrom<serde_json::Value> for VerifiableCredentialRecord {
...
+              "type": [
+                "VerifiableCredential"
+              ],
+              "issuer": "did:web:demopidprovider.acc.credenco.com:did:b49bc37c-6e24-442a-9597-069cdbc946bd",
+              "credentialSubject": {
+                "sub": "did:jwk:eyJhbGciOiJFZERTQSIsImNydiI6IkVkMjU1MTkiLCJraWQiOiJTSm9vQnQ4ZUtXUjhYejdxNEQyV0VkMmVTZ05xaTh0TTFnNGk4NGJQUVNjIiwia3R5IjoiT0tQIiwieCI6ImN0TTFkOThXTlRlZnN3LVBqbXpDWWFPZkFsWllULUlKZVFWY2JEaWtjWjAifQ",
+                "id": "urn:uuid:d0696c60-34bd-42ec-b1db-86f1a68e9487",
+                "iss": "did:web:demopidprovider.acc.credenco.com:did:b49bc37c-6e24-442a-9597-069cdbc946bd",
+                "nbf": 1741117731,
+                "exp": 1772653731,
```
Oran-Dan: is it me or do these claims usually live outside the `credentialSubject`, on the root level?

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -97,37 +113,120 @@ pub async fn handle_oid4vp_authorization_request(state: AppState, action: Action
...
+                    "No credentials available to fulfill the request".to_string(),
+                ));
+            }
+
+            // If multiple presentations are provided, this means that the `vp_token` in the Authorization
+            // Response will be a sequence which cannot be serialized into a x-www-form-urlencoded string by `reqwest`.
+            // See: https://github.com/nox/serde_urlencoded/issues/75#issuecomment-648257888
+            (sd_jwt_vc_count, 0) if sd_jwt_vc_count > 1 => {
+                return Err(AppError::Error(
+                    "Sending multiple presentations is not supported".to_string(),
```
Oran-Dan: It seems that sending multiple `jwt_vc_json` presentations is possible, so I would say this error message is not very accurate.

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -97,37 +113,120 @@ pub async fn handle_oid4vp_authorization_request(state: AppState, action: Action
...
+                ))
+            }
+            (sd_jwt_vc_count, jwt_vc_json_count) if (sd_jwt_vc_count > 0 && jwt_vc_json_count > 0) => {
+                return Err(AppError::Error(
+                    "Sending multiple presentations is not supported".to_string(),
```
Oran-Dan: see comment above

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -97,37 +113,120 @@ pub async fn handle_oid4vp_authorization_request(state: AppState, action: Action
...
+
+                oid4vp::AuthorizationResponseInput {
+                    verifiable_presentation_input: PresentationInputType::Presentation(
+                        presentation_builder.build().map_err(PresentationBuilderError)?,
+                    ),
+                    presentation_submission,
+                }
+            }
+            _ => {
+                unreachable!("Invalid combination of SD-JWT VC and JWT VC JSONs");
```
Oran-Dan: It seems this arm can only contain the possibility where ```(sd_jwt > 1, jwt_vc == 0)``` so also here I think the error message can be more explicit.

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
```diff
@@ -224,6 +323,7 @@ pub fn get_oid4vp_client_name_and_logo_uri(
                 .get(&ClaimFormatDesignation::JwtVcJson)
                 .and_then(|claim_format_property| match claim_format_property {
                     ClaimFormatProperty::Alg(alg) => alg.first().cloned(),
+                    ClaimFormatProperty::SdJwt { sd_jwt_alg_values, .. } => sd_jwt_alg_values.first().cloned(),
```
Oran-Dan: Shouldn't this iterate until we find an alg we support instead of only picking the first?

identity-wallet/src/state/credentials/mod.rs
nanderstabel: No, because all the data is contained in `credential_subject`.

This code basically decodes the complete SD-JWT and moves it into `credential_subject`:
```rust
                let credential_subject = sd_jwt_vc
                    .clone()
                    .into_disclosed_object(&Sha256Hasher::new())
                    .map_err(|_| AppError::Error("Failed to convert SD JWT VC to Disclosed Object".to_string()))?;
```

The frontend filters out all SD-JWT fields that are not meant to be displayed:
```typescript
    // TODO: to be removed once the backend properly propagates the to be displayed fields.
    // SD-JWT claims
    'iss',
    'nbf',
    'exp',
    'status',
    'iat',
    'sub',
    '_sd_alg',
    'cnf',
    'vct',
```

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: `jwt_vc_json` is a Credential not a Presentation. So yes multiple `jwt_vc_json` Credentials are possible as long as they're contained in one single Presentation

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: Sending multiple Credentials in one single Presentation is possible, sending multiple Presentations is not supported. (`vp_token` can be either a string or a vector of strings, acoording to spec, but we only support it being one single string)

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: It's logically impossible to reach this code, your example would already match on this arm: `(sd_jwt_vc_count, 0) if
sd_jwt_vc_count > 1 =>`.

But still even though it is impossible to reach this arm, it's better practice to still return an error instead of using
`unreachable!` so I will update :+1:

identity-wallet/src/state/credentials/mod.rs
nanderstabel: This is a direct result of this:
```rust
// TODO: Remove this workaround that is basically a way of disguising the SD JWT VC as a VC so that
// it can be displayed in the Frontend.
```

But perhaps what I could/should do is removing those fields directly [here](https://github.com/impierce/identity-wallet/pull/387/files/d9b5bb703b957b830c6bac8a546e9954ff871bfb#r2001478086) instead of filtering them out in the frontend :+1:

identity-wallet/src/state/credentials/mod.rs
nanderstabel: Will fix :+1:

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: Yes you're right I should definitely fix that!

identity-wallet/src/state/credentials/mod.rs
nanderstabel: I just added [this commit](https://github.com/impierce/identity-wallet/pull/387/commits/79701981007e5f34ec0a7ec5262c9d89f7fe99a8) which filters the SD-JWT specific claims in the backend instead of the frontend :+1:

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
Oran-Dan: Oke I see, and a "jwt_vc_json" and an "sd_jwt_vs" cannot be compiled into a single presentation like multiple "jwt_vc_json"s can?

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
Oran-Dan: But sd_jwt_vc is also a credential like jwt_vs_json right? or is that where my misunderstanding lies? Because multiple "jwt_vc"s are allowed but multiple "SD_jwt_vc"s are not.

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: Yeah that's right. I think maybe it is because VPs containing VCs is a W3C DM kind of thing, where the Credential is never altered and therefore the VP must contain the Holder's signature to ensure keybinding.

Inherent to SD-JWT credentials is that the Holder's keybinding can be appended to the credential JWT itself and therefore there is no need to be 'encapsulated' by an outer Presentation.

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: Yeah that's indeed confusing.. [here](https://github.com/impierce/identity-wallet/pull/387#discussion_r2001901418) I described why I think it is specified this way :+1:

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
nanderstabel: Fixed

Oran-Dan (dismissed):

Oran-Dan (approved):

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -54,45 +59,90 @@ impl TryFrom<serde_json::Value> for VerifiableCredentialRecord {
...
+                info!("Verifiable Credential parsed as a SD-JWT VC");
+
+                let issuance_date = sd_jwt_vc.claims().iat.map(|iat| iat.to_rfc3339()).unwrap_or_default();
+                let mut credential_subject = sd_jwt_vc
+                    .clone()
+                    .into_disclosed_object(&Sha256Hasher::new())
+                    .map_err(|_| AppError::Error("Failed to convert SD JWT VC to Disclosed Object".to_string()))?;
+
+                // Remove the SD-JWT specific fields that should not be displayed in the frontend.
+                for key in ["iss", "nbf", "exp", "status", "iat", "sub", "_sd_alg", "cnf", "vct"] {
```
daniel-mader, line 76: Could it lead to unexpected behavior if all of these fields are removed (even for "duplicate detection in demo cases")? What if one of my short-lived credentials gets prolonged by the issuer where only the `nbf`, `exp` and `iat` would be different? Would the "new one" then show up or the old one or actually both? If unsure, please add a TODO and open an issue for tracking.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -54,45 +59,90 @@ impl TryFrom<serde_json::Value> for VerifiableCredentialRecord {
...
+                // TODO: We are using this hash as Credential ID so that we can prevent credential duplication in
+                // demo situations. Now we can actually delete Credentials in UniMe we don't need to use the hash of the
+                // credential as the ID anymore. We should simply generate a random UUID.
+                // Derive the hash from the credential display.
+                let hash = {
+                    let type_value = credential_display["type"].clone();
+
+                    let mut credential_subject_value = credential_display["credentialSubject"].clone();
+
+                    // TODO(ngdil): Remove this hard-coded logic.
```
daniel-mader, line 110: Can this be removed now?

identity-wallet/src/state/credentials/mod.rs
nanderstabel: The new one would result in the same hash as the old one, so it will overwrite the old one. 

I don't think we have an explicit strategy for 'prolonging' credentials right? And since we now 'support' deletion of Credentials maybe we should simply stop using the Hash of the credentials in order to resolve?:
* `// TODO: We are using this hash as Credential ID so that we can prevent credential duplication in
                // demo situations. Now we can actually delete Credentials in UniMe we don't need to use the hash of the
                // credential as the ID anymore. We should simply generate a random UUID.`, and
* `// TODO(ngdil): Remove this hard-coded logic.`

identity-wallet/src/state/credentials/mod.rs
nanderstabel: Not in this current form, but we could simply start using random UUIDs instead of generating them from the Hash wyt? (see comment above)

daniel-mader: I meant if we can get rid of popping the `"Passport Number"`, `"Staff Number"` etc. out of the credential, since this code got touched in this PR as well. But I guess we make it a seperate issue to remove all hard-coded demo logic.

identity-wallet/src/state/credentials/mod.rs
daniel-mader: Okay, let's postpone until we actually introduce UUIDs fully.

daniel-mader (approved):

PR #414, opened 2024-11-22
fix: support `StatusList2021Entry` credentials, fetch logo from `.well-known/openid-credential-issuer`

Oran-Dan:
# Description of change
Previously the StatusList2021Entry type credential broke with our code, this is now fixed.
Also the metadata from the well-known is now more thoroughly searched to find the right logo for the credential.
Finally, the error messages in `validate_domain_linkage.rs` have been made more detailed.

- https://github.com/impierce/identity-wallet/issues/436

- Go to https://omzetbelasting.bd.demo.sphereon.com/
- Create a Omzetbelasting credential by filling in some dummy information
- Click continue and a screen with a qr code will appear
- Below it enter the cuddly-curufin URL
- Now it will continue to a JSON page, copy the `received_offer_id` at the top of the JSON payload on this page.
- Open Postman, make sure your environment variables are set to cuddly-curufin, go to POST `offers_accept`, fill in the `received_offer_id` you just copied and click SEND.
- Now it should return a 200 OK, scroll down to copy the `holder_credential_id` .
- paste this into the POST `presentations` request and click SEND
- Now copy the `presentation_id` and paste this into the `services/linked-vp` and click SEND.
- Go to `siopv2_authorization_requests` and click SEND, copy the QR code string
- Spin up Unime in dev mode, go to "Scan", and paste and process the QR code string
- You should now see the Omzetbelasting credential with logo


identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -189,15 +193,25 @@ async fn get_validated_linked_credential_data(
...
+            if validated_linked_domains.is_empty() {
+                if let Some(did_web_url) = extract_url_from_did_web(&issuer_did) {
+                    validated_linked_domains.insert(0, did_web_url);
+                }
+            }
+
             if !validated_linked_domains.is_empty() {
                 let validator = JwtCredentialValidator::with_signature_verifier(Verifier);
 
+                let options = JwtCredentialValidationOptions::new().status_check(StatusCheck::SkipUnsupported);
```
nanderstabel, line 209: I think it would be good to add a comment here to explain why the status check can be skipped

identity-wallet/resources/default_trust_list.json
```diff
@@ -1,5 +1,12 @@
 {
   "id": "b01f4a74-3005-4749-a030-c5444bc4dab5",
   "display_name": "Impierce Demos",
-  "entries": ["https://ngdil.com", "https://selv.iota.org", "https://thuiswinkel-agent.dev.impierce.com"]
+  "entries": [
+    "https://agent-dev.impierce.com",
```
daniel-mader: This was removed on purpose in a previous change, because it's outdated.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -189,15 +193,25 @@ async fn get_validated_linked_credential_data(
             info!("Issuer linked domains: {issuer_linked_domains:#?}");
 
             // Only linked verifiable credentials with at least one successful domain linkage validation are considered
-            let validated_linked_domains = get_validated_linked_domains(&issuer_linked_domains, &issuer_did).await;
+            let mut validated_linked_domains = get_validated_linked_domains(&issuer_linked_domains, &issuer_did).await;
+
+            // This is a fallback to get the url from a did:web to validate domain linkage. This is useful for companies who haven't implemented domain linkage yet.
```
daniel-mader: Please make sure to mark it with a TODO if this is temporary or should be re-evalutated later. If it is intended to stay in the logic, you don't need a TODO.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -208,19 +222,23 @@ async fn get_validated_linked_credential_data(
...
 
-                    OptionFuture::from(credential_subject.map(|credential_subject| async {
+                    if let Some(credential_subject) = credential_subject {
                         let name = get_name(credential_subject);
-                        let logo_uri = get_logo_uri(credential_subject).await;
+                        let logo_uri = get_logo_uri(credential_subject, &linked_verifiable_credential, &validated_linked_domains).await;
                         let issuance_date = linked_verifiable_credential.credential.issuance_date.to_rfc3339();
 
-                        LinkedVerifiableCredentialData {
+                        info!("Linkedverifiablecredential: name: {name:?}, logo_uri: {logo_uri:?}, issuance_date: {issuance_date}");
```
daniel-mader: This reads a bit weird: `Linkedverifiablecredential`. Consider Pascal casing or change to just `Linked VP:` _(or similar)_

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -208,19 +222,23 @@ async fn get_validated_linked_credential_data(
...
                             logo_uri,
                             issuance_date,
                             issuer_linked_domains: validated_linked_domains,
-                        }
-                    }))
-                    .await
+                        })
+                    }
+                    else {
+                        warn!("Failed to get credential_subject: {linked_verifiable_credential:#?}");
```
daniel-mader: Is this message consistent? What is actually logged here? The `credential_subject` or the `linked_verifiable_credential`?

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,134 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
         .properties
         .get("name")
         .or_else(|| credential_subject.properties.get("naam")) // TODO: "naam" is expected to be used in Dutch credentials
+        .or_else(|| credential_subject.properties.get("legal_person_name"))
```
daniel-mader: Is this a standardized claim name? If yes, please link the according spec somewhere.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,134 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
-                    &hash(image),
-                )
-                .await;
-                Some(image.to_string())
-            }),
-    )
+/// First try to get the logo URI from the credential subject.
+/// Then, if no success, iterate through the validated linked domains and try to fetch it from the well-known/openid-credential-issuer endpoint.
+/// In this endpoint, first we look inside the Display field, at the root.
+/// Then, if no success, we look inside the Credential Configuration Supported and try the match keys in there with the strings in the type array of the credential, in reverse order.
```
daniel-mader: I like the detailled description here!
I had a bit of trouble reading this part:
**[...] and try the match keys in there with the strings in the type array [...]**
Maybe it helps if you wrap actual field names in backticks to indicate a technical term, such as:
**[...] try to match a credential \`type\` (in reverse order).**

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,134 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+            }
+            Err(parse_err) => {
+                // Log parse error if the URI is invalid
+                warn!("Failed to parse logo URI: {:#?}, {:#?}", logo_uri_str, parse_err);
+            }
+        }
+    } else {
+        // Logo URI needs to be Some, even if it's value is not working, for the frontend to display the fallback icon.
+        // Otherwise, it doesn't display any icon.
+        // TODO: this needs to be fixed in the frontend.
```
daniel-mader: If it's a frontend bug, then don't try to "fix" it here by giving it a non-sense URI. I suggest getting rid of the entire `else` branch.

daniel-mader (changes requested):

identity-wallet/resources/default_trust_list.json
Oran-Dan: removed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: added TODO

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: Fixed, converted to Pascal

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed. it was consistent but clarified the text now.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed

M-Adam-Hus:
[fix: support StatusList2021Entry credentials and improve logo fetching from well-known/openid-credential-issuer](https://trello.com/c/oniYlU0c/71-fix-support-statuslist2021entry-credentials-and-improve-logo-fetching-from-well-known-openid-credential-issuer)

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
-            }),
-    )
+/// First, try to get the logo URI from the credential subject.
+/// If this doesn't succeed, iterate through the validated linked domains and try to fetch it from the well-known/openid-credential-issuer endpoint.
+/// In this endpoint, first we look inside the Display field, at the root.
+/// If we can't find a logo there, we look inside the Credential Configurations Supported field at the root.
+/// We try to match keys inside the Credential Configurations Supported object against the credential `type` array of the linked verifiable credential, in reverse order.
+/// At first success the loop breaks and we download the image.
+/// Otherwise, we use a fallback icon.
+async fn get_logo_uri(
```
nanderstabel, line 355: General remark: This function has gotten quite bloated and although we don't need to over-optimize this (since we need to come up with a non-custom solution anyway), it is still a good idea to make sure the current code is less bloated, easier to read and of course we need to make sure nothing unexpected can happen.

I will leave several comments throughout the function to highlight things that I think can cause problems and/or could be improved. 

In the final overall comment on this function I will share how I would address these problems altogether (this doesn't mean my suggestion should replace yours, it is just to illustrate an alternative. You can re-implement the code to your liking if you agree with the points I will make).

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+    // Check if logo URI was retrieved, else attempt to retrieve from well-known endpoint
+    if logo_uri.is_none() {
+        warn!("Failed to download logo URI from linked verifiable credential: {linked_verifiable_credential:#?}");
+        for domain in validated_linked_domains.iter() {
+            let mut well_known_endpoint = format!("{}.well-known/openid-credential-issuer", domain);
+            info!("Trying to fetch image from {well_known_endpoint} endpoint");
+            if let Ok(response) = reqwest::Client::new().get(&well_known_endpoint).send().await {
+                if let Ok(metadata) = response.json::<CredentialIssuerMetadata>().await {
+                    if let Some(display) = metadata.display {
+                        if let Some(image) = display[0].get("logo") {
```
nanderstabel: Indexing like this on a vector or array is dangerous because it will panic if the `display` vector is empty. In the future we should make sure cargo clippy catches this for us using [this lint](https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#indexing_slicing).

The following is a safe way to obtain the image here:
```rust
display.first().and_then(|display| display.get("logo"))
```

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+                }
+            }
+        }
+    }
+
+    if let Some(ref logo_uri_str) = logo_uri {
+        info!("Logo URI: {}", logo_uri_str);
+
+        // Parse the logo URI
+        match logo_uri_str.parse() {
```
nanderstabel, line 409: Two flaws I currently see with this code block though is the following situations:
1.
  * Let's say the value of `logo_uri` is `"not a URI"`?
  * This code block would detect that the value is not a valid `Url` and will therefore make the correct decision not to download the asset.
  * Still, this function will return `Some("not a URI")` and since it is not a `None` value, the frontend will think the backend actually found and persisted a logo.

So since you already validated that the logo could not be downloaded, it would be wise to return `None` if that is the case :)

2. If a `logo_uri` was found in line 359, then it will be downloaded in line 361, and then again in this code block (if `logo_uri` is a valid `Url`).

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
nanderstabel: This may differ person to person, but deeply nested `if let ...` sequences like this make code harder to follow in my opinion. 

I personally prefer using `.and_then()` and `.map()` (functional style) where I can because I think they're easier to follow.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+                                    let display = &credential_conf.display;
+                                    if let Some(image) = display[0].get("logo") {
+                                        if let Some(image_url) = image.get("url") {
+                                            logo_uri = image_url.as_str().map(ToString::to_string);
+                                            break;
+                                        }
+                                    }
+                                }
+                            }
+                        }
```
nanderstabel: The logic in both match cases here is basically the same except that in the second case multiple types are being checked. This unnecessary code duplication makes it harder to see what is going on and can be easily prevented.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+                        OneOrMany::Many(types) => {
+                            for type_ in types {
+                                info!("Trying to fetch from Credential Configuration Supported: {}", type_);
+                                if let Some(credential_conf) = credential_confs_supported.get(type_) {
+                                    let display = &credential_conf.display;
+                                    if let Some(image) = display[0].get("logo") {
+                                        if let Some(image_url) = image.get("url") {
+                                            logo_uri = image_url.as_str().map(ToString::to_string);
+                                            break;
```
nanderstabel: While your other two `break` statements will break out of the outer for loop in line 378, this one here will only break out of the for loop in line 416 instead (which I don't think was your intention.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+                if let Ok(metadata) = response.json::<CredentialIssuerMetadata>().await {
+                    let credential_confs_supported = metadata.credential_configurations_supported;
+                    match &linked_verifiable_credential.credential.types {
+                        OneOrMany::One(type_) => {
+                            info!("Trying to fetch from Credential Configuration Supported: {}", type_);
+                            if let Some(credential_conf) = credential_confs_supported.get(type_) {
+                                let display = &credential_conf.display;
+                                if let Some(image) = display[0].get("logo") {
+                                    if let Some(image_url) = image.get("url") {
+                                        logo_uri = image_url.as_str().map(ToString::to_string);
```
nanderstabel: This pattern of extracting the `logo_uri` from the `display` property is quite common so perhaps deserves its own separate function.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
```diff
@@ -319,30 +337,135 @@ fn get_name(credential_subject: &Subject) -> Option<String> {
...
+            }
+            Err(parse_err) => {
+                // Log parse error if the URI is invalid
+                warn!("Failed to parse logo URI: {:#?}, {:#?}", logo_uri_str, parse_err);
+            }
+        }
+    }
+
+    logo_uri
+}
```
nanderstabel: My complete refactor suggestion:

```rust
async fn get_logo_uri(
    credential_subject: &Subject,
    linked_verifiable_credential: &DecodedJwtCredential<Value>,
    validated_linked_domains: &[Url],
) -> Option<String> {
    let mut logo_uri = credential_subject
        .properties
        .get("image")
        .and_then(Value::as_str)
        .map(ToString::to_string);

    // Check if logo URI was retrieved, else attempt to retrieve from well-known endpoint
    if logo_uri.is_none() {
        for domain in validated_linked_domains.iter() {
            let well_known_endpoint = format!("{}.well-known/openid-credential-issuer", domain);
            info!("Trying to fetch image from {well_known_endpoint} endpoint");
            if let Ok(response) = reqwest::Client::new().get(&well_known_endpoint).send().await {
                if let Ok(metadata) = response.json::<CredentialIssuerMetadata>().await {
                    logo_uri = metadata.display.as_deref().and_then(extract_logo_uri_from_display);

                    if logo_uri.is_some() {
                        break;
                    }
                }
            }
            // TODO: Due to mixing 2 specs here, the oid4vci and linked verifiable presentation spec, we lose the Credential Issuer Identifier (CII) during the linked vp flow.
            // The CII tells us where exactly we can add "/.well-known/openid-credential-issuer" to fetch the Credential Issuer Metadata, in which we might find the logo.
            // For now we assume it's the same domain as the linked domain.
            // But this is no guarantee and the code below is one such workaround.
            let well_known_endpoint = format!("{}oid4vci/.well-known/openid-credential-issuer", domain);
            info!("Trying to fetch image from {well_known_endpoint} endpoint");
            if let Ok(response) = reqwest::Client::new().get(&well_known_endpoint).send().await {
                if let Ok(metadata) = response.json::<CredentialIssuerMetadata>().await {
                    logo_uri = linked_verifiable_credential.credential.types.iter().find_map(|type_| {
                        info!("Trying to fetch from Credential Configuration Supported: {}", type_);
                        metadata
                            .credential_configurations_supported
                            .get(type_)
                            .map(|credential_configuration| credential_configuration.display.as_ref())
                            .and_then(extract_logo_uri_from_display)
                    });

                    if logo_uri.is_some() {
                        break;
                    }
                }
            }
        }
    }

    OptionFuture::from(logo_uri.map(|logo_uri| async move {
        info!("Logo URI: `{}`", logo_uri);

        let _ = download_asset(
            logo_uri
                .parse()
                .inspect_err(|parse_err| warn!("Failed to parse logo URI: `{}`, {}", logo_uri, parse_err))
                .ok()?,
            &hash(&logo_uri),
        )
        .await
        .inspect_err(|err| warn!("Failed to download logo URI: {}", err));
        Some(logo_uri.to_string())
    }))
    .await
    .flatten()
}

fn extract_logo_uri_from_display(display: &[Value]) -> Option<String> {
    display
        .first()
        .and_then(|display| display.get("logo"))
        .and_then(|logo| logo.get("uri").or(logo.get("url")))
        .and_then(|url| url.as_str())
        .map(ToString::to_string)
}
```

nanderstabel (changes requested):

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: good point, fixed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: True! we should add these lints asap

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: great points, thanks a lot

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: agreed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: done

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: Yes, I ended up with your suggestion being the cleanest

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: Also used your suggestion here with the `find_map`, nice function otherwise I would have used 2 breaks in a row.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: So most of your suggestions I agree with and more or less end up with then trying myself. Only the asset downloading I've mostly kept the same as before.

nanderstabel (approved):

daniel-mader (approved):

PR #488, opened 2025-02-14
feat: store app password in key store of device, unlock with biometrics

### LINKED ISSUE #68: Biometric unlock
### Description

Unlock the secure storage using biometric methods supported by the user device

### Motivation

Improved security, enhanced convenience

### Requirements

1. detect available biometrics options, immediately ask the user for authentication
2. when authentication unsuccessful, show password input field
3. biometrics can be enabled/disabled in settings

### Open Questions

_No response_

### Are you planning to contribute this in a PR?

Yes

daniel-mader:
# Description of change
The user can opt-in to use biometrics to unlock the app.

- closes #68 
- depends on
  - https://github.com/impierce/identity-wallet/pull/364
  - https://github.com/impierce/identity-wallet/pull/498
  - https://github.com/impierce/identity-wallet/pull/541


nanderstabel:
I've looked into the failing tests and there are 2.5 causes:
1.
`test_qr_code_scanned_handle_siopv2_authorization_request` and `test_qr_code_scanned_handle_oid4vp_authorization_request` fail because during both tests, the Authorization Response is send to http://example.com, however this domain doesn't exist anymore and therefore the tests both time out. So these failing tests are not introduced in this PR. I've validated that this was the issue by replacing the domain to http://example.org (http://example.test also doesn't exist anymore), you can check out the changes in the `fix/example-org` branch. This would be a quick fix, but relying on external domain is not a good idea like we've discussed before. Perhaps it is better to change the test domains to https://www.impierce.com/ (because this domain will live forever :wink:) since the only requirement for the tests is that it can send an HTTP request (they don't care about the responses, only whether the domain exists). Of course ultimately the best solution would be to test against a test UniCore instance. 

Edit: it can be simply solved by skipping sending the Authorization Response during tests:
```rust
    #[cfg(not(feature = "test_utils"))]
    if provider_manager.send_response(&response).await.is_err() {
        info!("failed to send response");
        return Err(SendAuthorizationResponseError);
    }
```

in both `handle_siopv2_authorization_request.rs` and `handle_oid4vp_authorization_request.rs` 

2.
`test_get_state_create_new` and `test_locale_stays_unchanged_on_profile_creation` are failing because their corresponding fixture JSON files do not contain the newly introduced `biometrics_enabled` field in the `CreateNew` Action. So you can either add those fields to the corresponding fixtures, or probably better: add `#[serde(default)]` so that the missing field will default to `false`.
```rust
pub struct CreateNew {
    pub name: String,
    pub picture: String,
    pub theme: AppTheme,
    pub password: String,
    #[serde(default)]
    pub biometrics_enabled: bool,
}
```

2.5
`verifier_successfully_verifies_eddsa_signed_data` is still flaky because `test_credential_is_removed_from_appstate_and_from_stronghold_and_image_is_deleted` is not serial. This can easily be fixed by adding the `serial` attribute:
```rust
    #[tokio::test]
    #[serial_test::serial]
    async fn test_credential_is_removed_from_appstate_and_from_stronghold_and_image_is_deleted()
```

identity-wallet/src/state/common/reducers/unlock_storage.rs
```diff
@@ -13,9 +13,41 @@ use oid4vc::oid4vci::Wallet;
...
+                        let mut debug_messages = state.debug_messages.clone();
+                        debug_messages.push_back("Wrong Stronghold password".to_string());
+                        debug_messages
+                    },
+                    current_user_prompt: None,
+                    ..state
+                });
+            }
+        }
+
```
nanderstabel: This seems to me a complete separate functionality  from fully unlocking the storage or is it not?  Depending on `check_password_only`, either the password is checked OR the storage is fully unlocked. 

Wouldn't it make more sense then to create a `CheckPassword` Action + corresponding reducer for this?

identity-wallet/src/migrations/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# App State and Data Migrations
+
+Since UniMe is not semantically versioned, the following overview of migrations can be helpful for debugging and testing backwards compatibility.
+
+| UniMe Version | `AppState` Version | Git commit | Description                                |
+| ------------- | -----------------: | ---------- | ------------------------------------------ |
+| `0.8.2`       |                `1` | `ae1b555`  | Initial version, introduce `version` field |
+
+In order to get hold of a serialized version of the `AppState` at a specific time, simply checkout the source code at the corresponding version or git commit hash, load Ferri's profile and inspect the `state.json` in the data dir specific to your system.
```
nanderstabel: typo: `Ferri's`

unime/src/i18n/en/index.ts
```diff
@@ -216,6 +222,30 @@ By using our Service, you acknowledge that you have read, understood, and agree
...
+            TOUCH_ID: 'fingerprint',
+          },
+          IOS: {
+            FACE_ID: 'Face ID',
+            TOUCH_ID: 'Touch ID',
+          },
+          GENERIC: 'biometrics',
+        },
+        ENABLE: {
+          DIALOG_TITLE: 'Enable {type:string}',
```
nanderstabel, line 241: Wouldn't `Enable login with {type:string}` be slightly better?

unime/src/i18n/en/index.ts
```diff
@@ -216,6 +222,30 @@ By using our Service, you acknowledge that you have read, understood, and agree
...
+          DIALOG_CONTENT: 'Please enter your password to enable {type:string}.',
+        },
+        DISABLE: {
+          DIALOG_TITLE: 'Disable {type:string}',
```
nanderstabel, line 245: `Enable login with {type:string}`?

unime/src/i18n/nl-NL/index.ts
```diff
@@ -199,6 +205,30 @@ Door onze Service te gebruiken, erken je dat je deze Voorwaarden heeft gelezen,
...
+            TOUCH_ID: 'vingerafdruk',
+          },
+          IOS: {
+            FACE_ID: 'Face ID',
+            TOUCH_ID: 'Touch ID',
+          },
+          GENERIC: 'biometrie',
+        },
+        ENABLE: {
+          DIALOG_TITLE: '{type} inschakelen',
```
nanderstabel, line 224: `Inloggen met {type} inschakelen`?

Or in case you want to keep it the way it is, then make sure to capitalize:
* `FACE_ID: 'gezichtsherkenning'` -> `FACE_ID: 'Gezichtsherkenning'`
* `TOUCH_ID: 'vingerafdruk'` -> `FACE_ID: 'Vingerafdruk'`

The latter option would cause problems though in `DIALOG_CONTENT: 'Voer je wachtwoord in om {type} in te schakelen'` and `DIALOG_CONTENT: 'Voer je wachtwoord in om {type} uit te schakelen'` I suppose since `{type}` is in the middle of those sentences..

unime/src/i18n/nl-NL/index.ts
nanderstabel, line 1: In general in the Dutch translations, sometimes `{type}` occurs at the start of the phrase/sentence so in some cases `type` needs to be capitalized and in other cases not. I guess that's probably a bit tricky to generalize that over all the different languages? 

You could use a workaround by ensuring that `{type}` will never be at the start of the sentence and

nanderstabel (commented):
Rust code looks pretty solid, nicely done! Just left some questions and suggestions regarding the translations.

identity-wallet/src/migrations/CHANGELOG.md
maiertech: Maybe also mention where the data dirs are located on desktop for Mac and Linux.

maiertech (dismissed):

identity-wallet/src/state/common/reducers/unlock_storage.rs
daniel-mader: Yeah, I agree. I initially thought it would be a "lightweight" unlock, but it's actually two different things. 👍

identity-wallet/src/migrations/CHANGELOG.md
daniel-mader: Addressed both comments ✅

unime/src/i18n/en/index.ts
daniel-mader: That's too long ... :/

unime/src/i18n/en/index.ts
daniel-mader: Do you mean "Disable"?

unime/src/i18n/nl-NL/index.ts
daniel-mader: It is actually capitalize in the frontend due to that very specific issue ... So only on one page, the word is capitalized and in others it's not. I know this is not ideal, but we'll eventually move to a different i18n library anyway which hopefully solves the issue.

unime/src/i18n/nl-NL/index.ts
daniel-mader: see above

daniel-mader:
@nanderstabel, regarding the failing tests:
1. Let's fix this in a separate PR since it seems like a larger fix. Related issues:
- #492
- https://github.com/impierce/identity-wallet/issues/495
2. fixed ✅ 
2.5 fixed ✅

identity-wallet/src/state/common/reducers/unlock_storage.rs
daniel-mader: Refactored the flag out into its own action 👍

unime/src/i18n/nl-NL/index.ts
nanderstabel: Alright then let's leave it the way it is for now :+1:

nanderstabel (approved):

Issue #466, opened 2025-02-04
Implement data migration strategy


Oran-Dan:
### Description

Add data migration functionality.

### Motivation

Necessary to keep UniMe usage consistent between version upgrades.

### Requirements

1. Add `version` field to the `AppState`
2. Load stored `AppState` json file into json `Value`
3. Execute migration functions as required (by comparing versions)
4. Deserialize into `AppState`
5. Write back migrated `AppState`

### Are you planning to contribute this in a PR?

Yes

PR #498, opened 2025-02-23
feat: introduce versioning of `AppState`, prepare data migration path

Oran-Dan:
# Description of change
I've added a simple unsigned integer version field to the AppState.
Data migration functions will be ordered by incrementing version in a Vector.
The version field will then be used every time when loading the AppState from the STATE_FILE to check whether an update and data migration needs to take place. 
The version field will then be the index in the data migration function Vector from which the function(s) will start to be applied.
After updating and migrating the state is immediately saved again.

- #466

Added unit tests.
Existing tests pass.


identity-wallet/src/state/mod.rs
```diff
@@ -78,6 +78,7 @@ impl AppStateContainer {
 
 pub const SUPPORTED_SIGNING_ALGORITHMS: &[Algorithm] = &[Algorithm::EdDSA, Algorithm::ES256];
 pub const SUPPORTED_DID_METHODS: &[&str] = &["did:jwk", "did:key"];
+pub const APPSTATE_VERSION: u64 = 1;
```
daniel-mader: I think we should call it `CURRENT_VERSION` (or probably just `VERSION`), since it already resides inside `src/state/mod.rs`, so it's clear that it tracks the state.

identity-wallet/src/state/mod.rs
```diff
@@ -86,6 +87,8 @@ pub const SUPPORTED_DID_METHODS: &[&str] = &["did:jwk", "did:key"];
 #[ts(export)]
 #[serde(default)]
 pub struct AppState {
+    /// This field will be used for Data Migration between version updates.
```
daniel-mader: I don't think this comment adds any more meaning, because having a `version` field inside a `struct` already carries the meaning enough.

identity-wallet/src/persistence.rs
```diff
@@ -74,7 +95,7 @@ pub async fn save_state(app_state: &AppState) -> anyhow::Result<()> {
     // Here we take out the credentials field before saving the state,
     // being sensitive data they should only be stored in the stronghold, nowhere else.
     let mut json_app_state = serde_json::to_value(app_state)?;
-    json_app_state["credentials"] = serde_json::Value::Array(Vec::new());
+    json_app_state["credentials"] = Value::Array(Vec::new());
```
daniel-mader: I always prefer to keep the explicit `serde_json` in front of an arbitrary data type such as `Value` or `Object`, so you instantly know where this comes from without having to look at the imports first.

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +220,90 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
 pub fn hash(url: &str) -> String {
     sha256::digest(url).to_string()
 }
+
+// Data migration functions
+
+/// This function is used to migrate the app state from one version to the next.
+pub async fn appstate_version_update(
+    mut app_state_object: Map<String, Value>,
+    version: u64,
```
daniel-mader: What does this `version` do? Is this the current version or the target version?

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +220,90 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
 pub fn hash(url: &str) -> String {
     sha256::digest(url).to_string()
 }
+
+// Data migration functions
+
+/// This function is used to migrate the app state from one version to the next.
+pub async fn appstate_version_update(
```
daniel-mader: I believe that function naming conventions always start with a verb, so `state_update()` should rather be `update_state()`. Also, we're technically not "updating" anything, but the term commonly used when changing a data model is "migrating". You could call it `apply_state_migrations()`.

identity-wallet/src/state/mod.rs
```diff
@@ -86,6 +87,8 @@ pub const SUPPORTED_DID_METHODS: &[&str] = &["did:jwk", "did:key"];
...
+    pub version: u64,
```
daniel-mader: Couldn't we safely use a smaller size here, such as `u32`? Even if we ever reach the limit of a `u32` (highly unlikely), we can still switch to a `u64`. I believe this would make the produced TypeScript bindings a bit easier, since it currently produces a `version: bigint`, whereas it should produce a simpler `version: number` data type when you pick a smaller type in Rust.

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +220,90 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+    mut app_state_object: Map<String, Value>,
+    version: u64,
+) -> anyhow::Result<AppState> {
+    // Ordered vector of all migration functions up to the current version.
+    let migrations: Vec<fn(&mut Map<String, Value>)> = vec![];
+
+    for (index, migration) in migrations.iter().enumerate() {
```
daniel-mader: I think implying the "target version" directly from the index of the migrations vector is highly dangerous. An explicit mapping of the version number to the migration function would be less error-prone.
You could do something like _(pseudo code)_:
```
// read "version" field or default to "1".
let mut version = (json.get("version") as u32).unwrap_or(1);

// apply migration function if "version" is smaller than desired version
while version < CURRENT_VERSION {
        json = match version {
            1 => migrate_v1_to_v2(json),
            2 => migrate_v2_to_v3(json),
            _ => throw error: "No migration defined for version: {}",
        };
// update the "version" from within the loop from the newly parsed json
        version = json.get("version") as u32;
    }
```

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +63,27 @@ pub async fn load_state() -> anyhow::Result<AppState> {
     let state_file = STATE_FILE.lock().unwrap().clone();
     let bytes = read(state_file).await?;
     let content = String::from_utf8(bytes)?;
-    let app_state: AppState = serde_json::from_str(&content)?;
-    debug!("state loaded from disk");
+
+    // Load state to a serde_json::Object instead of direct deserialization to avoid different versions breaking.
```
daniel-mader: I feel like "avoid breaking" is a bit too vague.
What about something like this?
```
// Load state into a `serde_json::Object` first to run data model migrations before deserialization into `AppState`.
```

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +63,27 @@ pub async fn load_state() -> anyhow::Result<AppState> {
...
+    let app_state_object: Map<String, Value> = serde_json::from_str(&content)?;
+    let version = app_state_object.get("version").unwrap().as_u64().unwrap(); // TODO: remove unwrap
```
daniel-mader: I guess the `unwrap()` is okay if you change it into a `unwrap_or(1)` to make "1" the default version.

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +63,27 @@ pub async fn load_state() -> anyhow::Result<AppState> {
...
+            let app_state = serde_json::from_str(&content)?;
+            debug!("state loaded from disk");
+            app_state
+        }
+        _ => {
+            debug!(
+                "state version mismatch, performing data migration and appstate update from {} to {}",
+                version, APPSTATE_VERSION
+            );
+            let app_state = appstate_version_update(app_state_object, version).await?;
```
daniel-mader: Encapsulating the migrations into another function which then calls the individual migration functions doesn't really add to the readability. I think the migration functions could be called here directly, such as `migrate_v1_to_v2(app_state_object)`. Then you could also drop the `version` from the function signature.

identity-wallet/src/persistence.rs
daniel-mader: What do you think, @nanderstabel?

identity-wallet/src/state/mod.rs
Oran-Dan: I would actually disagree: there is also a "unime" version and "identity-wallet" version inside the toml files and this version pertains only to the AppState and is decoupled from any other version we have inside the codebase. Since it truly pertains to the AppState only, it should be stated as such I think

identity-wallet/src/state/mod.rs
Oran-Dan: resolved

identity-wallet/src/persistence.rs
Oran-Dan: Makes sense, it is quite a vague type name without the serde_json. I've added them again.

identity-wallet/src/persistence.rs
Oran-Dan: resolved

identity-wallet/src/persistence.rs
Oran-Dan: changed the name to `outdated_version`

identity-wallet/src/state/mod.rs
Oran-Dan: refactored to u32.

identity-wallet/src/persistence.rs
Oran-Dan: resolved

identity-wallet/src/persistence.rs
Oran-Dan: yes you're right, done

identity-wallet/src/persistence.rs
Oran-Dan: I agree this is less error prone. I'll refactor

identity-wallet/src/persistence.rs
Oran-Dan: I am much more of a split each operation into a separate function kinda guy. To me it adds a lot to readibility, grouping code together under specific names and variables, like paragraphs. But i'm fine either way.

identity-wallet/src/state/mod.rs
daniel-mader: Okay, agreed. But then let's be consistent with the naming and call it `APP_STATE_VERSION`, since you also use `AppState` and `app_state`.

identity-wallet/src/persistence.rs
daniel-mader: I don't think there such a thing as an "outdated version". There can only be a current version and a target version (to be migrated to). After a migration step has completed successfully, the "current version" is bumped until the target version has been reached.

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +226,96 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+    }
+
+    let app_state_value = serde_json::Value::Object(app_state_object);
+    let app_state: AppState = serde_json::from_value(app_state_value)?;
+
+    save_state(&app_state)
+        .await
+        .map_err(|_| AppError::Error("Failed to save state after applying migrations".to_string()))?;
+
+    info!("state successfully migrated, AppState: {:?}.", app_state);
```
daniel-mader: This would be logged, even if the state is already at the latest version and there are no migrations applied. I think we can safely remove it, since successful migrations are already logged inside the `while` loop.

identity-wallet/src/state/mod.rs
Oran-Dan: Done

identity-wallet/src/persistence.rs
Oran-Dan: Okay, renamed it to `current_version`

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +226,96 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+        outdated_version = app_state_object
+            .get("version")
+            .and_then(|v| v.as_u64())
+            .map(|v| v as u32)
+            .ok_or_else(|| AppError::Error("Failed to get version while migrating AppState".to_string()))?;
+
+        info!("state successfully migrated to AppState version {}.", outdated_version);
+    }
+
+    let app_state_value = serde_json::Value::Object(app_state_object);
```
daniel-mader: Is there any benefit in having a stronger type over a simple `serde_json::Value` here?

```
app_state_value = serde_json::Value::Object(app_state_object);
```

Since `serde_json::Value` already represents any valid JSON value and we handle it gracefully anyway, I see no benefit in defining it as a `serde_json::Map<>` (although it's the expected type of the `state.json` file). We could completely get rid of this conversion just to make the compiler happy.

identity-wallet/src/persistence.rs
daniel-mader: Okay, let's keep it in a separate function then.

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +62,34 @@ pub async fn load_state() -> anyhow::Result<AppState> {
...
+
+    // Load state into a `serde_json::Object` first to run data model migrations before deserialization into `AppState`.
+    let app_state_object: serde_json::Map<String, serde_json::Value> = serde_json::from_str(&content)?;
+    // Get the version from the state file, if no version field is found, default to 1.
+    let version = app_state_object
+        .get("version")
+        .unwrap_or(&serde_json::Value::from(1))
+        .as_u64()
+        .map(|v| v as u32)
+        .ok_or_else(|| AppError::Error("Failed to get valid version while loading AppState".to_string()))?;
```
daniel-mader: This error message gives too little detail about what went wrong exactly. We should define a new error variant in `AppError`, such as `StateMigrationError` which has parameters for "current version" and "target version" and some text field that will give detailled information about why an app state migration could not be performed.

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +62,34 @@ pub async fn load_state() -> anyhow::Result<AppState> {
...
+
+    let app_state: AppState = match version {
+        APP_STATE_VERSION => {
+            let app_state = serde_json::from_str(&content)?;
+            debug!("state loaded from disk");
+            app_state
+        }
+        _ => {
+            debug!(
+                "state version mismatch, performing data migration and appstate update from {} to {}",
```
daniel-mader: This debug message does not really add any DX value since the migrations themselves also log each step.

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +62,34 @@ pub async fn load_state() -> anyhow::Result<AppState> {
...
+                version, APP_STATE_VERSION
+            );
+            let app_state = apply_state_migrations(app_state_object, version).await?;
+            debug!("state successfully loaded from disk and migrated");
```
daniel-mader: I feel this debug message is in the wrong place. It should be logged when the `state.json` has been loaded from disk as a `serde_json` value.

The desired log is this:

```
state loaded from disk (version 1)
applying migration 1 --> 2 ...
applying migration 2 --> 3 ...
successful migration
persisted state back to disk (version 3)
...
log in ferris
```

On next run:

```
state loaded from disk (version 3)
no migrations required
...
log in ferris
```

identity-wallet/src/persistence.rs
Oran-Dan: ok, removed it

identity-wallet/src/persistence.rs
Oran-Dan: This is done to allow for easy mutations. Plenty of `Map` associated functions to `insert()`/`append()` etc. which are not available for `Value`. The benefits will present themselves when we start making actual migration functions :)

identity-wallet/src/persistence.rs
Oran-Dan: Agreed, done

identity-wallet/src/persistence.rs
Oran-Dan: removed

identity-wallet/src/persistence.rs
Oran-Dan: implemented

identity-wallet/src/persistence.rs
Oran-Dan: removed

identity-wallet/src/persistence.rs
Oran-Dan: I've made the error messages more explicit

identity-wallet/src/error.rs
```diff
@@ -96,6 +96,8 @@ pub enum AppError {
     StateFileDeletionError(#[source] anyhow::Error),
     #[error("Failed to find TrustList with ID `{0}`")]
     TrustListNotFoundError(String),
+    #[error("Failed to migrate AppState version `{0}` to latest version `{1}`: {2}")]
```
daniel-mader: What if the version in the json file on disk is `"version": 2` and the app has the `APP_STATE_VERSION = 5`, then an error in the migration step from 3 => 4 would result in the log:

```
Failed to migrate AppState version `3` to latest version `4`: Could not apply logic xyz
```

==> This error could also be thrown when it's not the "latest" version.

identity-wallet/src/persistence.rs
daniel-mader: Okay, sounds like we might benefit from stricter typing then. 👍

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +230,114 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+        // This is a temporary solution to avoid infinite loops.
+        current_version += 1;
+
+        // current_version = app_state_object
+        //     .get("version")
+        //     .and_then(|v| v.as_u64())
+        //     .map(|v| v as u32)
+        //     .ok_or_else(|| AppError::Error("Failed to get version while migrating AppState".to_string()))?;
+
+        info!(
```
daniel-mader: This could be a `debug!()` message to be consistent with the other migration logs.

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +230,114 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+            "state successfully migrated AppState version from {} to {}.",
+            previous_version, current_version
+        );
+    }
+
+    let app_state_value = serde_json::Value::Object(app_state_object);
+    let app_state: AppState = serde_json::from_value(app_state_value)?;
+
+    save_state(&app_state)
```
daniel-mader: Refactoring suggestion: the function wouldn't need to be `async` if it returned the `app_state` without calling `save_state()` itself. It would also be a lot easier to test since nothing needs to be persisted in a test, but only the return value of `apply_state_migrations()` would need to be asserted.

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +230,114 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+                "type": "redirect",
+                "target": "welcome"
+            }
+        }
+        "#,
+        )
+        .unwrap();
+        let const_app_state_str = serde_json::to_string(&const_app_state).unwrap();
+
+        assert_eq!(test_app_state_str, const_app_state_str);
```
daniel-mader: Comparing the string representation of two json objects is highly discouraged since keys could be ordered differently, while still representing the same underlying json object. You should compare two `serde_json::Value`s instead.

Oran-Dan: yes good point, fixed it

identity-wallet/src/persistence.rs
Oran-Dan: fixed

identity-wallet/src/error.rs
Oran-Dan: removed "latest"

identity-wallet/src/persistence.rs
Oran-Dan: Good point, refactored

identity-wallet/src/persistence.rs
nanderstabel: I'd suggest to use even stricter typing, since `serde_json::Value::Object` is basically like Typescripts `any` equivalent in Rust. If we're going to do the actual migrations on the level of `serde_json::Value::Object`, then the migration flow will always look something like this:
state_file --> Object --> Object updates --> AppState version 2
Which is doable, but it is a bit more error prone than a flow like this where we deserialize to an AppState version as soon as possible, and then do the migrations directly between AppState versions (because we can rely on Rust strong type system):
state_file --> AppState version 1 --> AppState version 2

I played around with this some time ago and this is what I came up with (it was just an experiment so not a perfect solution):
```rust
mod migration_idea {
    use serde::{Deserialize, Serialize};
    use std::collections::HashMap;

    #[derive(Serialize, Deserialize, Debug)]
    struct AppStateV1 {
        foo: String,
        bar: u32,
    }

    impl Into<AppStateV2> for AppStateV1 {
        fn into(self) -> AppStateV2 {
            AppStateV2 {
                foo: self.foo,
                // Update the type of the `bar` field
                bar: self.bar > 0,
            }
        }
    }

    #[derive(Serialize, Deserialize, Debug)]
    struct AppStateV2 {
        foo: String,
        bar: bool,
    }

    impl Into<AppStateV3> for AppStateV2 {
        fn into(self) -> AppStateV3 {
            AppStateV3 {
                foo: self.foo,
                bar: self.bar,
                // A new field is added
                new_feature: HashMap::new(),
            }
        }
    }

    #[derive(Serialize, Deserialize, Debug)]
    struct AppStateV3 {
        foo: String,
        bar: bool,
        new_feature: HashMap<String, String>,
    }

    impl Into<AppStateV4> for AppStateV3 {
        fn into(self) -> AppStateV4 {
            AppStateV4 {
                version: "v4".to_string(),
                foo: self.foo,
                bar: self.bar,
                // Renamed/replaced an already existing field
                renamed: self.new_feature,
            }
        }
    }

    #[derive(Serialize, Deserialize, Debug)]
    struct AppStateV4 {
        version: String,
        foo: String,
        bar: bool,
        renamed: HashMap<String, String>,
    }

    #[derive(Serialize, Deserialize, Debug)]
    #[serde(tag = "version", rename_all = "snake_case")]
    enum AppStateVersion {
        V1(AppStateV1),
        V2(AppStateV2),
        V3(AppStateV3),
    }

    impl AppStateVersion {
        fn migrate(self) -> Self {
            match self {
                AppStateVersion::V1(v1) => AppStateVersion::V2(v1.into()),
                AppStateVersion::V2(v2) => AppStateVersion::V3(v2.into()),
                v3 @ AppStateVersion::V3(_) => v3,
            }
        }
    }

    impl From<AppStateVersion> for AppStateV4 {
        fn from(mut old: AppStateVersion) -> Self {
            while !matches!(old, AppStateVersion::V3(_)) {
                old = old.migrate();
            }
            match old {
                AppStateVersion::V3(v3) => v3.into(),
                _ => unreachable!(),
            }
        }
    }

    type AppState = AppStateV4;

    #[test]
    fn migration_from_version_1() {
        let state_file = serde_json::json!({
            "version": "v1",
            "foo": "hello",
            "bar": 42,
        });

        let app_state: AppState = serde_json::from_value::<AppStateVersion>(state_file.clone())
            .map(Into::into)
            .or(serde_json::from_value::<AppState>(state_file))
            .unwrap();

        println!("app_state: {:#?}", app_state)
    }

    #[test]
    fn migration_from_version_2() {
        let state_file = serde_json::json!({
            "version": "v2",
            "foo": "hello",
            "bar": true,
        });

        let app_state: AppState = serde_json::from_value::<AppStateVersion>(state_file.clone())
            .map(Into::into)
            .or(serde_json::from_value::<AppState>(state_file))
            .unwrap();

        println!("app_state: {:#?}", app_state)
    }

    #[test]
    fn migration_from_version_3() {
        let state_file = serde_json::json!({
            "version": "v3",
            "foo": "hello",
            "bar": true,
            "new_feature": {
                "key": "value"
            }
        });

        let app_state: AppState = serde_json::from_value::<AppStateVersion>(state_file.clone())
            .map(Into::into)
            .or(serde_json::from_value::<AppState>(state_file))
            .unwrap();

        println!("app_state: {:#?}", app_state)
    }

    #[test]
    fn from_version_4() {
        let state_file = serde_json::json!({
            "version": "v4",
            "foo": "hello",
            "bar": true,
            "renamed": {
                "key": "value"
            }
        });

        let app_state: AppState = serde_json::from_value::<AppStateVersion>(state_file.clone())
            .map(Into::into)
            .or(serde_json::from_value::<AppState>(state_file))
            .unwrap();

        println!("app_state: {:#?}", app_state)
    }
```

Oran-Dan: Well this was a thought as well. But as you can see from the example already this would require really a lot more code to do the job. To the extend that I think it outweighs the benefits. Especially the structs for example will most of the time be 97% the same. Therefore a lot of duplicate code will exist, making it hard to navigate this code. (Comments could alleviate this pain tho).

nanderstabel: Ok yeah makes sense :ok_hand:

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +62,43 @@ pub async fn load_state() -> anyhow::Result<AppState> {
...
+    let version = app_state_object
+        .get("version")
+        .unwrap_or(&serde_json::Value::from(1))
+        .as_u64()
+        .map(|v| v as u32)
+        .ok_or_else(|| {
+            AppError::AppStateMigrationError(
+                0,
+                APP_STATE_VERSION,
+                "Failed to get valid current version from AppState loaded from disk".to_string(),
```
nanderstabel, line 86: This will only be returned if the `app_state_object` **does** contain a `version` field, but it is not of type `u64`. I wonder if we could simply assume that this would never happen, and if it does happen, then we ignore it and simply set/update the `version` field once the migrations have been successful. In that case we could simplify this block to something like:
```rust
    let version = app_state_object
        .get("version")
        .and_then(serde_json::Value::as_u64)
        .map(|v| v as u32)
        .unwrap_or(0);
```

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +235,110 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+pub fn apply_state_migrations(
+    app_state_object: serde_json::Map<String, serde_json::Value>,
+    mut current_version: u32,
+) -> anyhow::Result<AppState, AppError> {
+    while current_version < APP_STATE_VERSION {
+        // this code is commented out because we don't have any migrations yet.
+        // match current_version {
+        //     1 => migrate_v1_to_v2(&mut app_state_object)?,
+        //     _ => return Err(AppError::AppStateMigrationError(current_version, APP_STATE_VERSION, "Unsupported AppState version")),
+        // }
```
nanderstabel: Is this true? Can't we already create a `migrate_v0_to_v1` migration function?

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +62,43 @@ pub async fn load_state() -> anyhow::Result<AppState> {
     let state_file = STATE_FILE.lock().unwrap().clone();
     let bytes = read(state_file).await?;
     let content = String::from_utf8(bytes)?;
-    let app_state: AppState = serde_json::from_str(&content)?;
+
+    // Load state into a `serde_json::Object` first to run data model migrations before deserialization into `AppState`.
+    let app_state_object: serde_json::Map<String, serde_json::Value> = serde_json::from_str(&content)?;
     debug!("state loaded from disk");
+
+    // Get the version from the state file, if no version field is found, default to 1.
```
nanderstabel: Shouldn't this default to 0?

identity-wallet/src/persistence.rs
nanderstabel: Since this is an update to the  `AppState` schema right?:
```rust
pub struct AppState {
    pub version: u32, // <-- update
```

So currently all UniMe versions have version `0` (implicitly) and therefore their `app_state` should be migrated to version `1` (explicitly)

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +235,110 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+
+        let app_state_value = Value::Object(app_state_object);
+        let app_state: AppState = serde_json::from_value(app_state_value)?;
+
+        Ok(app_state)
+    }
+
+    fn dummy_migrate_v0_to_v1(app_state_object: &mut Map<String, Value>) {
+        app_state_object.insert("version".to_string(), Value::from(1));
+    }
```
nanderstabel: If you agree with [my comment above](https://github.com/impierce/identity-wallet/pull/498/files#r1975571962), then there is no need really for these dummy functions since you can just test `apply_state_migrations` directly.

identity-wallet/src/persistence.rs
```diff
@@ -199,3 +235,110 @@ pub fn persist_asset(file_name: &str, id: &str) -> Result<(), AppError> {
...
+        //     .ok_or_else(|| AppError::Error("Failed to get version while migrating AppState".to_string()))?;
+
+        debug!(
+            "state successfully migrated AppState version from {} to {}.",
+            previous_version, current_version
+        );
+    }
+
+    let app_state_value = serde_json::Value::Object(app_state_object);
+    let app_state: AppState = serde_json::from_value(app_state_value)?;
```
nanderstabel: So here is one thing related to [my earlier comment](https://github.com/impierce/identity-wallet/pull/498/files#r1974707545) that I still would like to point out:

In this line we completely trust that we've applied all the migrations properly, so we are confident that we can now deserialize `app_state_value` into the current `AppState` version.

But, what if we just made **one** single mistake during one of the migrations? E.g. some fields type change from `i32` to `u32` and we forget to check for negative numbers during a migration: 
```rust

struct PreviousAppState {
    some_field: i32
}

struct CurrentAppState {
    some_field: u32
}
```

Now this is a very simple example and it perhaps seems unrealistic or very difficult to miss. But the `AppState` could become a very deeply nested structure, where some nested fields could even be types defined in third-party libraries that we don't have control over. If at some point we update one of our dependencies which contains a schema update that we are not aware of, then we're basically already screwed:

Of course we want to assume that would never be the case because we will be very thorough in reviewing and testing our migration functions. But it could still happen and the result would be quite a disaster:

This:
```rust
let app_state: AppState = serde_json::from_value(app_state_value)?;
```
would return an error from `fn apply_state_migrations` to `fn load_state` which would propagate the error up to the `fn get_state` reducer here:
```rust
let mut state = load_state().await.unwrap_or_default();
```

where `unwrap_or_default()` will assign a default value to `state` meaning that all the users data is now gone due to one single migration error.

So I am not saying we should go one way or the other (although I have my preference for utilizing Rusts type system), but we need to fully accept the risks that come with sidelining Rusts type system.

identity-wallet/src/persistence.rs
```diff
@@ -61,8 +62,43 @@ pub async fn load_state() -> anyhow::Result<AppState> {
     let state_file = STATE_FILE.lock().unwrap().clone();
     let bytes = read(state_file).await?;
     let content = String::from_utf8(bytes)?;
-    let app_state: AppState = serde_json::from_str(&content)?;
+
+    // Load state into a `serde_json::Object` first to run data model migrations before deserialization into `AppState`.
+    let app_state_object: serde_json::Map<String, serde_json::Value> = serde_json::from_str(&content)?;
```
nanderstabel, line 73: So the loaded state here doe not contain any `credentials` at this point, since they are stored in the stronghold file and only added to the `AppState` after the `unlock_storage` reducer did its thing. 

However, the schema for how we store Credentials in Stronghold could change too and therefore we need a migration strategy for that as well.

identity-wallet/src/persistence.rs
Oran-Dan: Yeah very tiny probability but I chose this option because in my mind I don't see defaulting to 0 and then applying all migrations on a corrupted AppState as logical.

identity-wallet/src/persistence.rs
Oran-Dan: We could do this but i'm not sure if this is error prone since V0 is like a random box of different versions. One with no trustlists yet, the other with a different Connections struct etc.

identity-wallet/src/persistence.rs
Oran-Dan: This ties into wether we start from V0 or V1

identity-wallet/src/persistence.rs
Oran-Dan: oh good catch! In my mind that means we need to change that unwrap_or_default() line of code more than anything else though. Because I do think 100% of changes in migrations simply need to be tested. 
The hazard of dep updates is a risk even without migrations and perhaps something we should discuss separately?

identity-wallet/src/persistence.rs
Oran-Dan: Good point!

identity-wallet/src/persistence.rs
daniel-mader: I agree with @Oran-Dan on this one. While it is the safest bet to have all versions of the AppState as structs, it adds significantly to the codebase. In your examples, you only use 2-3 fields per version, but we're already at 11-12 and most of them are heavily nested as well. Operating on a more loosely typed `serde_json::Value::Object` is a conscious decision here which allows us to only keep one `struct AppState` (latest).
We absolutely got to have solid testing of the migration functions in place to counter the looser types though.

identity-wallet/src/persistence.rs
daniel-mader: I think @nanderstabel was mainly referring to testing here. I would advise we're not going to run any migrations when introducing this PR into the code base. You're right @Oran-Dan, there are different versions out there and considering them all as the same "version 0" is going to break some, but not all. Let's just not migrate anything and just accept that all will break. That's why we're in beta and not handling any production data yet. This will also be the last time this happens since we will always run migrations from now on.

identity-wallet/src/persistence.rs
nanderstabel: Edit: my comment here actually better fits the discussion [here](https://github.com/impierce/identity-wallet/pull/498#discussion_r2018702417) so I moved it :+1:

identity-wallet/src/persistence.rs
nanderstabel: --moved--

I think I would still default to 0 but ok.

So currently what would happen if the state object does not contain a value for `version`, you would default to 1, and since `APP_STATE_VERSION` == 1, it would match on the first match arm. Then you deserialize the original `content` into `AppState` but because the content still does not contain a `version` field it will default to `0`. So now you log that there are `no migrations needed` but the AppState now contains `version: 0` so that will be saved to to the state.

Then next time the user opens the app, the state needs to be loaded again. This time the state object does contain a value for `version` (which is `0`)  therefore it does not match on the first match arm anymore so `apply_state_migrations` is called. It doesn't contain any explicit migration functions so the `app_state_object` is not updated, still you log `state successfully migrated AppState version from 0 to 1.`.

Then you deserialize the `app_state_object` again into the `AppState`, but the `version` is still `0`. And this will repeat every time the state is loaded again and the reason for that is that you never migrate from version 0 to version 1.

I guess this behavior will stop once `APP_STATE_VERSION` is set to `2` because then you will return an error:
```
Failed to migrate AppState version `0` to version `2`: Unsupported AppState version`
```

So it would eventually resolve itself I guess so maybe this is the way you intended it. It just seemed a bit odd to me.

Because in my mind the above would so easily be resolved by actually migrating from version 0 tot 1:
```rust

// Data migration functions
fn migrate_v0_to_v1(app_state_object: &mut Map<String, Value>) {
    app_state_object.insert("version".to_string(), Value::from(1));
}

/// This function is used to migrate the app state from one version to the next.
pub fn apply_state_migrations(
    mut app_state_object: serde_json::Map<String, serde_json::Value>,
    mut current_version: u32,
) -> anyhow::Result<AppState, AppError> {
    while current_version < APP_STATE_VERSION {
        match current_version {
            0 => migrate_v0_to_v1(&mut app_state_object),
            _ => {
                return Err(AppError::AppStateMigrationError(
                    current_version,
                    APP_STATE_VERSION,
                    "Unsupported AppState version".to_string(),
                ))
            }
        }

        let previous_version = current_version;
        // TODO: This is a temporary solution to avoid infinite loops.
        current_version += 1;

        // current_version = app_state_object
        //     .get("version")
        //     .and_then(|v| v.as_u64())
        //     .map(|v| v as u32)
        //     .ok_or_else(|| AppError::Error("Failed to get version while migrating AppState".to_string()))?;

        debug!(
            "state successfully migrated AppState version from {} to {}.",
            previous_version, current_version
        );
    }

    let app_state_value = serde_json::Value::Object(app_state_object);
    let app_state: AppState = serde_json::from_value(app_state_value)?;

    Ok(app_state)
}
```

I could definitely be missing something crucial, just trying to understand why you would not want to migrate 0 to 1.

identity-wallet/src/persistence.rs
Oran-Dan: This is fixed

identity-wallet/src/data_migration.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    let app_state_value = serde_json::Value::Object(app_state_object);
+    // If somehow the AppState updates weren't succesful after all, the function will throw an error trying to deserialize here
+    let app_state: AppState = serde_json::from_value(app_state_value)?;
+
+    Ok(app_state)
+}
+
+/////     Data migration functions     /////
```
daniel-mader:
```
/////     Data migration functions     /////

                                       ^^^^^
```
Why dis?

identity-wallet/src/data_migration.rs
```diff
@@ -0,0 +1,83 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{data_migration::apply_state_migrations, state::AppState};
+    use serde_json::{Map, Value};
+
+    #[tokio::test]
+    async fn appstate_version_update_test() {
```
daniel-mader: Convention is to have the `test_` at the beginning:
```
appstate_version_update_test()    // current

test_appstate_version_update()    // conventional
```

identity-wallet/src/state/common/reducers/unlock_storage.rs
```diff
@@ -12,6 +12,7 @@ use oid4vc::oid4vc_manager::ProviderManager;
 use oid4vc::oid4vci::Wallet;
 use std::sync::Arc;
 
+// TODO: implement Data Migration regarding the credentials here
```
daniel-mader: Are you sure this is the right place for the TODO? If not sure, please remove it for now since we'll stumble on it anyway if we get to it.

unime/src/lib/stores.ts
```diff
@@ -3,12 +3,15 @@ import { writable } from 'svelte/store';
 // TODO: run some copy task instead of importing across root to make the frontend independent
 import type { AppState } from '@bindings/AppState';
 
+const APP_STATE_VERSION = 1;
```
daniel-mader: Does the frontend really need to be aware of the version as well? I think it's entirely backend-defined, so if we need a default value to satisfy the `type AppState`, please consider inlining `version: 0` in `empty_state` to avoid confusion.

identity-wallet/src/persistence.rs
```diff
@@ -61,17 +63,53 @@ pub fn initialize_storage(app_handle: &tauri::AppHandle) -> anyhow::Result<()> {
...
+                0,
+                APP_STATE_VERSION,
+                "Failed to get valid current version from AppState loaded from disk".to_string(),
+            )
+        })?;
+
+    let app_state: AppState = match version {
+        APP_STATE_VERSION => {
+            let app_state = serde_json::from_str(&content)?;
+            debug!("no migrations needed");
```
daniel-mader: We can be a little more verbose here, I think:
`App state is at current version 3, no migrations needed`

Same applies for the success message after all has been applied:
`App state successfully migrated to version 3 (latest)`

identity-wallet/src/persistence.rs
```diff
@@ -61,17 +63,53 @@ pub fn initialize_storage(app_handle: &tauri::AppHandle) -> anyhow::Result<()> {
...
-    let content = String::from_utf8(bytes)?;
-    let app_state: AppState = serde_json::from_str(&content)?;
+    let content = String::from_utf8(bytes.clone())
+        .map_err(|e| AppError::Error(format!("Failed to convert utf8, {:?}, to String: {}", bytes, e)))?;
+
+    // Load state into a `serde_json::Object` first to run data model migrations before deserialization into `AppState`.
+    let app_state_object: serde_json::Map<String, serde_json::Value> = serde_json::from_str(&content)?;
     debug!("state loaded from disk");
+
+    // Get the version from the state file, if no version field is found, default to 1.
```
nanderstabel: Small thing: 'default to 1' needs to be updated to `default to 0'

identity-wallet/src/data_migration.rs
Oran-Dan: fixed

identity-wallet/src/state/common/reducers/unlock_storage.rs
Oran-Dan: sure, makes sense

identity-wallet/src/data_migration.rs
Oran-Dan: fixed

identity-wallet/src/persistence.rs
Oran-Dan: fixed

unime/src/lib/stores.ts
Oran-Dan: sure, front-end is completely your call.
Fixed it as proposed

nanderstabel (approved):

daniel-mader (approved):

PR #511, opened 2025-03-03
fix: relative time calculations and tests

### LINKED ISSUE #510: Vitest unit tests break
Vitest unit tests break in CI. Figure out why and fix.

maiertech:
# Description of change

Two unit tests testing `formatRelativeDateTime` failed:

1. The "1 month ago" test came back as "4 weeks ago" instead of "1 month ago". This was due to February 2025 having only 28 days. This is correct and I fixed the unit test to allow two results.
2. The "2 months ago" test came back as "last month". This was due to a rounding error. I now use `Math.round` instead of `Math.floor`.

- Fixes #510.

Be sure to reference any related issues by adding `fixes issue #`.

Vitest unit tests shouls pass.

### DIFF
--- a/unime/src/lib/utils.ts
+++ b/unime/src/lib/utils.ts
@@ -83,19 +83,26 @@ export function formatDateTime(isoDate: string, locale: Locale, test = false) {
 }
 
 export function formatRelativeDateTime(isoDate: string, locale: Locale) {
-  // 1 min, 1 hour, 1 day, 1 week, 1 month, 1 year.today
+  const date = new Date(isoDate);
+  const now = new Date();
+
+  const diffInSeconds = (date.getTime() - now.getTime()) / 1000;
+
+  // Thresholds in seconds: 1 min, 1 hour, 1 day, 1 week, 1 month, 1 year.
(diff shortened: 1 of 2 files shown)


unime/src/lib/utils.ts
```diff
@@ -83,19 +83,26 @@ export function formatDateTime(isoDate: string, locale: Locale, test = false) {
 }
 
 export function formatRelativeDateTime(isoDate: string, locale: Locale) {
-  // 1 min, 1 hour, 1 day, 1 week, 1 month, 1 year.today
+  const date = new Date(isoDate);
+  const now = new Date();
+
+  const diffInSeconds = (date.getTime() - now.getTime()) / 1000;
```
daniel-mader, line 89: Is there a specific reason why you chose to flip it? Is there a benefit in doing it this way?
Before it was: `Date.now() - new Date(isoDate)` (we can expect now to be larger, since `isoDate` is in the past) which would result in a positive value.
Now it is: `new Date(isoDate) - new Date()` (inverted), that's why you now need `Math.abs()` as well.

maiertech: Now we can not only do relative dates into the past but also into the future.

daniel-mader (approved):

PR #516, opened 2025-03-11
feat: email verification flow

daniel-mader:
# Description of change

This introduces communication with an external "email verification service".

The following actions are triggered by the user and/or UniMe itself:

- ServiceHealthCheck
- SendVerificationEmail
- RedeemCode
- ResetEmailVerification

**Note for reviewers: there are potential changes not present in this PR, but other PRs stacked "on top".**

n/a

The latest public UniMe version (`0.9.7`) has been verified by multiple individual testers.


identity-wallet/src/state/verified_data/reducers/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        // TODO: handle error
+        let json_response: serde_json::Value = response.json().await.unwrap();
+        info!("[<<<] {}", json_response);
+        let id = json_response.get("id").unwrap().as_str().unwrap();
+        let expires_at = json_response.get("expires_at").unwrap().as_str().unwrap();
+        let validation_expiration_in_secs = json_response
+            .get("validation_expiration_in_secs")
+            .unwrap()
+            .as_i64()
+            .unwrap() as i32;
```
nanderstabel: These would all be deserialization/type errors so I'd suggest to create a type for the response and use something like this:
```rust
        let json_response: VerificationResponse =
            response.json().await.map_err(|err| AppError::Error(err.to_string()))?;
        info!("[<<<] {}", json_response);
        ...
```

identity-wallet/src/state/verified_data/reducers/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+                    AppState {
+                        verified_data: VerifiedData {
+                            email_verification: None,
+                        },
+                        ..state
+                    },
+                    std::sync::Arc::new(action),
+                )
+                .await
+                .unwrap());
```
nanderstabel: `.await.unwrap()` --> `.await?`

identity-wallet/src/state/verified_data/reducers/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+            .post(url)
+            .header("X-API-KEY", EMAIL_VERIFICATION_SERVICE_API_KEY)
+            .json(&body)
+            .send()
+            .await
+            .inspect_err(|err| {
+                warn!("Failed to send verification code: {}", err);
+            })
+            .ok()
+            .unwrap();
```
nanderstabel: This would just log a warning and then crash the app. Should we simply return the error like so?:

```rust
            .await
            .map_err(|err| AppError::Error(format!("Failed to send verification code: {err}")))?;
```

identity-wallet/src/state/verified_data/reducers/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+pub async fn redeem_code(state: AppState, action: Action) -> Result<AppState, AppError> {
+    if let Some(action) = listen::<RedeemCode>(action) {
+        let session_id = state
+            .verified_data
+            .email_verification
+            .as_ref()
+            .unwrap()
+            .verification_id
+            .as_ref()
+            .unwrap();
```
nanderstabel: Not sure if logically can never panic, but still it's probably better to add some error handling here:
```rust
        let session_id = state
            .verified_data
            .email_verification
            .as_ref()
            .and_then(|email_verification| email_verification.verification_id.clone())
            .ok_or(AppError::Error(
                "Tried to redeem a code without an active email verification flow".to_string(),
            ))?;
```

identity-wallet/src/state/verified_data/reducers/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+                .await
+                .unwrap());
+            }
+            _ => {
+                let error: serde_json::Value = response.json().await?;
+                warn!("Failed to redeem code: {}", error);
+                return Ok(AppState {
+                    verified_data: VerifiedData {
+                        email_verification: Some(EmailVerification {
+                            error: Some(error.get("error").unwrap().as_str().unwrap().to_string()),
```
nanderstabel: I guess this logically can not panic, but still let's return an error here instead?

unime/src/i18n/nl-NL/index.ts
```diff
@@ -401,6 +403,42 @@ Door onze Service te gebruiken, erken je dat je deze Voorwaarden heeft gelezen,
...
+        ITEM_2: {
+          TITLE: '',
+          DESCRIPTION: '',
+        },
+      },
+      ADD: {
+        NAVBAR_TITLE: 'E-mail toevoegen',
+        LABEL: 'Hoe wilt u dit e-mailadres noemen?',
+        LABEL_DISCLAIMER: 'Alleen zichtbaar voor jou',
+        LABEL_PLACEHOLDER: 'Persoonlijk e-mail',
```
nanderstabel: `Persoonlijk` --> `Persoonlijke`

identity-wallet/src/state/verified_data/actions/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+};
+use crate::state::{actions::ActionTrait, Reducer};
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+// Service health check
+
+#[derive(Serialize, Deserialize, Debug, Clone, TS)]
+#[ts(export, export_to = "bindings/actions/ServiceHealthCheck.ts")]
+pub struct ServiceHealthCheck {
```
nanderstabel: nit: with other actions in other files we use doc-comments so perhaps for consistency sake we should do that here as well

identity-wallet/src/state/verified_data/actions/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+impl ActionTrait for RedeemCode {
+    fn reducers<'a>(&self) -> Vec<Reducer<'a>> {
+        vec![reducer!(redeem_code)]
+    }
+}
+
+// Redeem code
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+// #[ts(export, export_to = "bindings/actions/ResetEmailVerification.ts")]
```
nanderstabel: Can be removed I guess

identity-wallet/src/state/verified_data/reducers/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+                warn!("Failed to send verification code: {}", err);
+            })
+            .ok()
+            .unwrap();
+
+        info!("[<<<] {:?}", response);
+
+        match response.status().as_u16() {
+            200 => {
+                let credential_offer_value: String = response.text().await.unwrap();
```
nanderstabel: `.await.unwrap()` --> `.await?`

nanderstabel (changes requested):
Just some minor comments but all the logic looks sound!

I also realized that this flow is essentially just an OID4VCI Issuance pre-authorized code flow with [tx_code](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-4.1.1-5.2.2.2.1) which means that in a next iteration of the email verification service we should probably make use of that. 

`tx_code` is also a DIIPv4 requirement so we need to enable support for it pretty soon anyway.

identity-wallet/src/state/verified_data/reducers/mod.rs
nanderstabel: Actually it can panic I think because the `error` value that is returned is not guaranteed to have an inner `error` field right? When the email verification service is down for instance we can not expect there to be an `error` field in the response or can we?

So either return an error here or default to `None`, e.g.:
```rust
error: error.get("error").and_then(Value::as_str).map(ToString::to_string)
```

daniel-mader: Yes, that's a lot safer and more idiomatic Rust! Thanks for the suggestion! 👍

nanderstabel (approved):

PR #553, opened 2025-05-20
feat: hide `Developer mode` setting by default

### LINKED ISSUE #542: Hide developer mode by default
### Description

The `Developer mode` menu item should be hidden by default and not presented to users. To reveal it, we can use a similar pattern as on Android, such as tapping the UniMe logo in `Settings > About UniMe` 7 times in a row.

### Motivation

Enhance end user experience, without limiting developer capabilities

### Resources

_No response_

### To-do List

- [ ] Hide `Developer mode` settings entry by default
- [ ] Count taps on UniMe logo in `About UniMe`
- [ ] Reveal `Developer mode` settings entry

daniel-mader:
# Description of change
The "Developer mode" setting is hidden by default to prevent distracting users.
It can be made visible by going to `Settings` > `About UniMe` and tapping the UniMe logo seven times.
This pattern is used by Android as well to enable developer mode on Android itself.

- closes #542
- depends on #516 

n/a


identity-wallet/src/state/dev_mode/reducers/show_setting.rs
```diff
@@ -0,0 +1,35 @@
...
+        actions::{listen, Action},
+        dev_mode::{actions::show_setting::ShowDevModeSetting, DevMode},
+        user_prompt::CurrentUserPrompt,
+        AppState,
+    },
+};
+
+use log::debug;
+
+pub async fn show_setting(mut state: AppState, action: Action) -> Result<AppState, AppError> {
```
nanderstabel: nit: conceptually the 'old'/current state should not be mutable. The reducer should just expect the current state as is and return a new state or return an error. You can rewrite the body of the outer `if` statement to something like this:
```rust
        if show {
            return Ok(AppState {
                current_user_prompt: Some(CurrentUserPrompt::Redirect {
                    target: "me/settings/dev_mode".to_string(),
                }),
                ..state
            });
        } else {
            return Ok(AppState {
                dev_mode: DevMode::Off,
                current_user_prompt: Some(CurrentUserPrompt::Redirect {
                    target: "me/settings/dev_mode".to_string(),
                }),
                ..state
            });
        }
```

Keeping the incoming current state immutable also prevents us in the future from accidentally updating the state where we shouldn't.

nanderstabel (approved):

Issue #579, opened 2025-07-23
Add `credentialStatus`


Oran-Dan:
### Description

Users will now be able to clearly see the status of their credentials and have it up to date every time they open the app.

### Motivation

CredentialStatus and revocation is a crucial part of the Digital Credentials ecosystem.

### Requirements

- [x]  (Backend) All credentials should be checked for their status and updated each time upon opening UniMe
- [x]  (Backend) Check status when receiving a credential, when invalid do not add the credential.
- [x]  (Backend) Create action and reducer to be dispatched for manual refresh of Credential Status.
- [x]  (Backend) add CredentialStatusData struct to DisplayCredential to reflect the status of the credential (Valid, Unvalid, Suspended) and the timestamp last checked.
- [ ]  (Backend) unable Invalid credentials from being shared; add logic to filter out Unvalid candidates for a DCQL request
- [ ]  (Frontend) grey out Invalid credentials, sort those credentials to the bottom of the list
- [ ]  (Frontend) create manual refresh option in main page all credentials overview
- [ ]  (Frontend) create manual refresh option in credential overview

### Are you planning to contribute this in a PR?

Yes

JelleMillenaar:
Please all check the status of a credential on delivery, we don't want to add revoked credentials to the wallet ;)

Oran-Dan:
We should also check the status again when clicking a specific credential overview. Users might actually not close the app for very long time periods as they're not of the habit of truly closing any apps. Furthermore, users might be waiting for a live update on status (most likely when hearing it will be unrevoked/unsuspended).

PR #497, opened 2025-02-21
feat: validate credential structure and data model

Oran-Dan:
# Description of change
Improves security and verification process. 

Blocked by this issue:
- https://github.com/iotaledger/identity.rs/issues/1550

Which has been solved in this PR:
- https://github.com/iotaledger/identity.rs/pull/1570

Now waiting for IOTA to release a new version of their crate which contains this fix.

## Links to any relevant issues
- #480

## How the change has been tested
Added unit tests.
Old tests pass.

## Definition of Done checklist
Add an `x` to the boxes that are relevant to your changes.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #585, opened 2025-08-11
feat: check and refresh credential status

Oran-Dan:
# Description of change

This PR adds the functionality of checking the credentialStatus, according to the IETF OAuth Token Status List spec https://datatracker.ietf.org/doc/draft-ietf-oauth-status-list/11/ 
In combination with the DIIPv4 Profile which outlines how to define the VC Data Model `credentialStatus` property for this spec.
The action-reducer pair "RefreshCredentialStatus" has been added.

Blocked by:
- #497

- #579

Integration tests have been added. No unit tests have been added since the function involves fetching, signing amongst others, making unit tests actually harder to code than just a full integration test.


identity-wallet/src/state/credentials/actions/refresh_credential_status.rs
```diff
@@ -0,0 +1,19 @@
...
+use crate::state::credentials::reducers::refresh_credential_status::refresh_credential_status;
+use crate::state::{actions::ActionTrait, Reducer};
+
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+#[derive(Serialize, Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/RefreshCredentialStatus.ts")]
+pub struct RefreshCredentialStatus {
+    pub credential_id: Option<String>,
```
daniel-mader: This could be just `id` since it's clear from the context that it is the id of the credential.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -31,6 +32,17 @@ pub struct DisplayCredential {
     pub connection_id: Option<String>,
     // TODO: should this be moved to `metadata`?
     pub display_name: String,
+    pub credential_status: Option<CredentialStatusData>,
```
daniel-mader: This could be just `status` since it's clear from the context that this is the status of the credential. Otherwise, we'll have something like `credential.credential_status` in the frontend (compared to `credential.status`).

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,148 @@
...
+        else {
+        }
+        return Ok(AppState { ..state });
+    }
+    Ok(state)
+}
+
+/// Represents the credential status as defined in the OAuth Token Status List specification and the DIIP profile.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TSLCredentialStatus {
```
daniel-mader: This casing is non-conventional. Either go `TslCredentialStatus` or leave out the `Tsl` altogether.

identity-wallet/src/state/credentials/actions/refresh_credential_status.rs
```diff
@@ -0,0 +1,19 @@
+use crate::reducer;
+use crate::state::credentials::reducers::refresh_credential_status::refresh_credential_status;
+use crate::state::{actions::ActionTrait, Reducer};
+
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+#[derive(Serialize, Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/RefreshCredentialStatus.ts")]
+pub struct RefreshCredentialStatus {
```
daniel-mader, line 10: Is there a difference between a "refresh" and a simple "fetch"? Does "refresh" imply there's an update to an existing status? Or would a more generic "FetchCredentialStatus" be more fitting? I'm fine with either, I'm just interested if there's a difference.

unime/src-tauri/tests/fixtures/actions/refresh_all_credential_statuses.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "[Credential] Refresh Status",
+  "payload": {}
```
daniel-mader: This cannot be triggered by the frontend right now, since the exported binding requires a `credential_id` to be present (it can still be `null`). Would that have the same effect to refresh all?

identity-wallet/bindings/actions/RefreshCredentialStatus.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+export interface RefreshCredentialStatus { credential_id: string | null, }
```
daniel-mader: I think `credential_id?: string` would be cleaner.

identity-wallet/bindings/credentials/DisplayCredential.ts
```diff
@@ -1,4 +1,5 @@
 // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
 import type { CredentialMetadata } from "./CredentialMetadata";
+import type { CredentialStatusData } from "./CredentialStatusData";
 
-export interface DisplayCredential { id: string, format: string, issuer_name: string, data: any, metadata: CredentialMetadata, connection_id?: string, display_name: string, }
\ No newline at end of file
+export interface DisplayCredential { id: string, format: string, issuer_name: string, data: any, metadata: CredentialMetadata, connection_id?: string, display_name: string, credential_status: CredentialStatusData | null, }
```
daniel-mader: `credential_status?: CredentialStatusData` would be more idiomatic.

daniel-mader: It would be even cleaner like this: `status?: CredentialStatus`

identity-wallet/bindings/actions/RefreshCredentialStatus.ts
daniel-mader: Or even omit the "credential" and just go `id?: string`.

identity-wallet/src/state/credentials/reducers/send_credential_request.rs
daniel-mader, line 1: All of those imports are unused.

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
daniel-mader: This struct is unused.

Oran-Dan:
3 important comments which are included in the code:
- // TODO: test for possible poor latency/performance due to many stronghold interactions when many credentials need to be refreshed.
I think this can best be tested in UniMe itself rather than with cargo test which will only give us a partial indication of the time it takes, namely the backend.
- // This error handling means we don't panic when the refresh_credential_status function fails. 
// Instead we don't bother the user with any of the errors and keep the old status and simply don't update it.
// However, this is also not ideal. TODO: how to handle a status that consistently fails to refresh?
How to handle the TODO?
-  // TODO: currently we set the credential status when a credential is received. We might want to inform the user of this before accepting the credential already

identity-wallet/src/state/credentials/actions/refresh_credential_status.rs
Oran-Dan: I've kept the term "fetch" for the subfunction that does exactly that since i think "fetch" is technically a more specific term. That's why this function got the distinct name "Refresh...".

identity-wallet/src/state/credentials/reducers/refresh_all_credential_statuses.rs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    drop(state_guard);
+
+    Ok(AppState {
+        credentials,
+        ..state.to_owned()
+    })
+}
```
nanderstabel: Would it make sense to reuse the `refresh_credential_status` reducer here for example like this?:
```rust
pub async fn refresh_all_credential_statuses(state: AppState, action: Action) -> Result<AppState, AppError> {
    if let Some(_refresh_all_credential_statuses) = listen::<RefreshAllCredentialStatuses>(action) {
        let credential_ids: Vec<String> = state
            .credentials
            .iter()
            .map(|credential| credential.id.clone())
            .collect();

        let mut state = state;

        for credential_id in credential_ids {
            state = refresh_credential_status(state, Arc::new(RefreshCredentialStatus { credential_id })).await?;
        }

        return Ok(AppState { ..state });
    }
    Ok(state)
}
```

identity-wallet/src/state/credentials/actions/refresh_all_credential_statuses.rs
```diff
@@ -0,0 +1,17 @@
...
+use crate::state::{actions::ActionTrait, Reducer};
+
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+#[derive(Serialize, Deserialize, Debug, TS, Clone)]
+#[ts(export, export_to = "bindings/actions/RefreshAllCredentialStatuses.ts")]
+pub struct RefreshAllCredentialStatuses {}
+
+#[typetag::serde(name = "[Credential] Refresh All Statuses")]
```
nanderstabel: In order to stay consistent with the other Actions, let's change the `name` to `"[Credential] Refresh all statuses"`

identity-wallet/src/state/credentials/actions/refresh_credential_status.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub credential_id: String,
+}
+
+#[typetag::serde(name = "[Credential] Refresh Status")]
```
nanderstabel: Similar to https://github.com/impierce/identity-wallet/pull/585/files#r2405465873, let's change this to `"[Credential] Refresh status"`

identity-wallet/bindings/credentials/CredentialStatusData.ts
nanderstabel, line 1: This is generated form a type that been removed again, so this file can also be removed :)

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+
+pub async fn refresh_credential_status(state: AppState, action: Action) -> Result<AppState, AppError> {
+    if let Some(refresh_credential_status) = listen::<RefreshCredentialStatus>(action) {
+        let state_guard = state.core_utils.managers.lock().await;
+        let mut credentials = state.credentials.clone();
+        let credential_id = refresh_credential_status.credential_id;
+
+        if let Some(credential) = credentials.iter_mut().find(|c| c.id == credential_id) {
+            if let Some(credential_status_data) = credential.credential_status.as_mut() {
+                match fetch_credential_status(credential_status_data, state_guard.identity_manager.as_ref().unwrap())
```
nanderstabel: Instead of unwrapping, I think we should use a variable for `identity_manager` that is initilialized like this above this `match` block:
```rust
                let identity_manager = state_guard
                    .identity_manager
                    .as_ref()
                    .ok_or(AppError::MissingManagerError("identity"))?;

                match fetch_credential_status(credential_status_data, identity_manager).await {
```

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+                    .await
+                {
+                    Ok(status) => {
+                        info!("Successfully fetched new credential status {status:?} for credential with id: `{credential_id}`. The old_status was: {:?}", credential_status_data.status);
+                        credential_status_data.last_checked = DateUtils::new_date_string();
+                        credential_status_data.status = status;
+
+                        // Update the credential in StrongHold
```
nanderstabel: nit: `StrongHold` --> `Stronghold`

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+    let res = client
+        .get(uri)
+        .header(header::ACCEPT, accept_header.to_string())
+        .send()
+        .await
+        .map_err(AppError::FetchCredentialListError)?;
+
+    if !res.status().is_success() {
+        return Err(AppError::FetchCredentialListError(res.error_for_status().unwrap_err()));
+    }
```
nanderstabel: I think this can be refactored to:
```rust
    let res = client
        .get(uri)
        .header(header::ACCEPT, accept_header.to_string())
        .send()
        .await
        .and_then(reqwest::Response::error_for_status)
        .map_err(AppError::FetchCredentialListError)?;
```

Basically what `error_for_status` does is wrapping the Response in a `Result` which can then directly be mapped to `FetchCredentialListError` if the StatusCode was in the Error range.

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+            Ok(decompressed)
+        }
+        _ => {
+            // If no gzip encoding, return the body as is.
+            let jwt_bytes = res.bytes().await.map_err(AppError::FetchCredentialListError)?;
+            let jwt_vec_u8 = jwt_bytes.to_vec();
+
+            Ok(String::from_utf8(jwt_vec_u8).map_err(|_| AppError::GetCredentialStatusError)?)
+        }
+    }
```
nanderstabel: Just leaving this as an alternative (but feel free to keep the current implementation):
```rust
    let is_gzipped = res
        .headers()
        .get(header::CONTENT_ENCODING)
        .map_or(false, |encoding| encoding == "gzip");

    let bytes = res.bytes().await.map_err(AppError::FetchCredentialListError)?;

    if is_gzipped {
        decompress_gzip(&bytes).map_err(|e| {
            warn!("Failed to decompress gzip status list: {}", e);
            AppError::GetCredentialStatusError
        })
    } else {
        String::from_utf8(bytes.to_vec()).map_err(|e| {
            warn!("Status list was not valid UTF-8: {}", e);
            AppError::GetCredentialStatusError
        })
    }
```

identity-wallet/src/state/credentials/reducers/self_issue_credential.rs
```diff
@@ -168,8 +168,6 @@ pub async fn self_issue_credential(state: AppState, action: Action) -> Result<Ap
 
         let key: Uuid = vcr.display_credential.id.parse().expect("invalid uuid");
 
-        // TODO: add schema VC 2 validation
```
nanderstabel, line 171: Was this intentionally removed?

identity-wallet/src/state/credentials/reducers/send_token_request.rs
```diff
@@ -305,6 +308,29 @@ pub async fn send_token_request(state: AppState, action: Action) -> Result<AppSt
...
             validate_credential_types(&verifiable_credential_record.verifiable_credential)?;
+            // The credential status is set only when the credential status claim/property can be found and is in OAuth TSL format.
+            // If setting the credential status fails we currently catch the error and simply set the credential status field to None.
+            // TODO: we might want to inform the user of this before accepting the credential already
+            match set_credential_status(
+                &mut verifiable_credential_record,
+                state_guard
+                    .identity_manager
+                    .as_ref()
+                    .ok_or(MissingManagerError("identity"))?,
```
nanderstabel: Earlier in this function we already validate that the `identity_manager` exists, so if you change:
```rust
        let wallet = &state_guard
            .identity_manager
            .as_ref()
            .ok_or(MissingManagerError("identity"))?
            .wallet;
```

to:
```rust
        let identity_manager = state_guard
            .identity_manager
            .as_ref()
            .ok_or(MissingManagerError("identity"))?;
        let wallet = &identity_manager.wallet;
```

then we can keep the call to `set_credential_status` a bit 'leaner':
```rust
            match set_credential_status(&mut verifiable_credential_record, identity_manager).await {
```

identity-wallet/src/state/credentials/reducers/send_token_request.rs
```diff
@@ -305,6 +308,29 @@ pub async fn send_token_request(state: AppState, action: Action) -> Result<AppSt
             let mut verifiable_credential_record = VerifiableCredentialRecord::try_new(credential, claims)?;
             // Validate the credential against its corresponding credential JSON Schema.
             validate_credential_types(&verifiable_credential_record.verifiable_credential)?;
+            // The credential status is set only when the credential status claim/property can be found and is in OAuth TSL format.
+            // If setting the credential status fails we currently catch the error and simply set the credential status field to None.
+            // TODO: we might want to inform the user of this before accepting the credential already
+            match set_credential_status(
```
nanderstabel: The current implementation of `set_credential_status` currently takes the responsibility of both _producing_ and _assigning_, so I think it would be more idiomatic to split those two responsibilities for example by doing something like this:
```rust
            verifiable_credential_record.display_credential.credential_status =
                match get_credential_status(&verifiable_credential_record.verifiable_credential, identity_manager) {
                    Ok(credential_status) => {
                        info!(
                            "Successfully set credential status for credential with id: `{}`",
                            verifiable_credential_record.display_credential.id
                        );
                        Some(credential_status)
                    }
                    Err(err) => {
                        warn!("Failed to get credential status: {err}");
                        None
                    }
                };
```
* `set_credential_status` --> `get_credential_status`
* pass (immutable) `&verifiable_credential_record.verifiable_credential` instead of a mutable reference to `verifiable_credential_record`

identity-wallet/src/state/credentials/reducers/send_token_request.rs
```diff
@@ -430,6 +458,71 @@ fn get_credential_display_name(
...
+            idx: credential_status_claim.referenced_status_list.idx,
+            uri: credential_status_claim.referenced_status_list.uri,
+            last_checked: String::new(),
+        };
+
+        let status = fetch_credential_status(&credential_status_data, identity_manager).await?;
+        credential_status_data.status = status;
+        credential_status_data.last_checked = DateUtils::new_date_string();
+
+        verifiable_credential_record.display_credential.credential_status = Some(credential_status_data);
```
nanderstabel: These `if` and `if else` block are largely the same, they only differ in how the `credential_status_claim` is obtained. So, I think this function should be refactored so that it:
1. first obtains the `credential_status_claim` value
2. then, creates the `credential_status_data`
3. returns the `credential_status_data` (see https://github.com/impierce/identity-wallet/pull/585/files#r2405772340)

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+                info!("No credentialStatus found for credential with id: `{credential_id}`");
+
+                return Err(AppError::GetCredentialStatusError);
+            }
+        } else {
+            // This should never happen, as the credential ID send by the frontend is supposed to be valid.
+            warn!("No credential found with id: `{credential_id}`");
+
+            return Err(AppError::GetCredentialStatusError);
+        }
```
nanderstabel: Due to the nested `if let `Some(...)` blocks, these error messages are detached from their corresponding 'source' functions so I would suggest to refactor this by executing all the steps one by one (not nested), for example:
```rust
        let credential = credentials.iter_mut().find(|c| c.id == credential_id).ok_or_else(|| {
            // Comment
            warn!("log message");
            AppError::GetCredentialStatusError
        })?;

        let credential_status_data = credential.credential_status.as_mut().ok_or_else(|| {
            // Comment
            warn!("log message");
            AppError::GetCredentialStatusError
        })?;

    etc...
```

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+                                .map_err(AppError::StrongholdInsertionError)?;
+                        }
+                    }
+                    Err(e) => {
+                        // This error handling means we don't panic when the refresh_credential_status function fails.
+                        // Instead we don't bother the user with any of the errors and keep the old status and simply don't update it.
+                        // However, this is also not ideal. TODO: how to handle a status that consistently fails to refresh?
+                        warn!("Failed to refresh credential status for credential with id: `{credential_id}`.\nThe old status remains unchanged: {:?}\nError: {e}", credential_status_data.status);
+
+                        return Err(e);
```
nanderstabel: Because of this:
> "... we don't bother the user with any of the errors ..."

this block should not: `return Err(e);` but instead it should return:

```rust

                        drop(state_guard);

                        return Ok(AppState { credentials, ..state });
```

if I'm not mistaken?

identity-wallet/src/state/actions.rs
```diff
@@ -134,6 +136,10 @@ mod bindings {
         UpdateCredentialMetadata { payload: UpdateCredentialMetadata },
         #[serde(rename = "[Credential] Delete")]
         DeleteCredential { payload: DeleteCredential },
+        #[serde(rename = "[Credential] Refresh Status")]
+        RefreshCredentialStatus { payload: RefreshCredentialStatus },
+        #[serde(rename = "[Credential] Refresh All Statuses")]
```
nanderstabel: See: https://github.com/impierce/identity-wallet/pull/585/files#r2405465873

identity-wallet/src/state/actions.rs
```diff
@@ -134,6 +136,10 @@ mod bindings {
         UpdateCredentialMetadata { payload: UpdateCredentialMetadata },
         #[serde(rename = "[Credential] Delete")]
         DeleteCredential { payload: DeleteCredential },
+        #[serde(rename = "[Credential] Refresh Status")]
```
nanderstabel: See: https://github.com/impierce/identity-wallet/pull/585/files#r2405467953

identity-wallet/src/error.rs
```diff
@@ -100,6 +106,8 @@ pub enum AppError {
     TrustListNotFoundError(String),
     #[error("Failed to migrate AppState version `{0}` to version `{1}`: {2}")]
     AppStateMigrationError(u32, u32, String),
+    #[error("Provided code is invalid")]
+    InvalidTransactionCode(#[source] anyhow::Error),
```
nanderstabel: This was probably not intended?

unime/src-tauri/Cargo.toml
```diff
@@ -25,6 +25,7 @@ identity-wallet = { path = "../../identity-wallet" }
 iota-sdk = { git = "https://github.com/iotaledger/iota", package = "iota-sdk", tag = "v1.6.1" }
 jni = "0.21"
 log.workspace = true
+oauth_tsl.workspace = true
```
nanderstabel: This dependency is only being used in tests so it should be moved down under `[dev-dependencies]`.

Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = ["identity-wallet", "unime/src-tauri"]
 tauri = { version = "2.8.2", features = ["protocol-asset", "rustls-tls"] }
 tauri-build = "2.4.0"
 
+oauth_tsl = { git = "https://github.com/impierce/oauth-token-status-list", rev = "133d224" }
```
nanderstabel: This should be in alphabetical order

nanderstabel (changes requested):

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
```diff
@@ -0,0 +1,194 @@
...
+        .subject
+        .public_key(&key_id)
+        .await
+        .map_err(|_| AppError::GetCredentialStatusError)?;
+    let decoding_key = match jwt_header.alg {
+        Algorithm::EdDSA => DecodingKey::from_ed_der(&public_key),
+        Algorithm::ES256 => DecodingKey::from_ec_der(&public_key),
+        _ => {
+            // This panic should never happen since we initialize our public keys and identity_manager ourselves.
+            panic!("Unsupported algorithm: {:?}", jwt_header.alg);
```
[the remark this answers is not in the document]
nanderstabel: The `jwt_header` comes from an external source, so we can not assume that the `alg` is always eddsa or es256.

identity-wallet/src/state/credentials/reducers/refresh_all_credential_statuses.rs
Oran-Dan: Yes!

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
Oran-Dan: oh yes thanks, totally missed that unwrap!

identity-wallet/src/error.rs
Oran-Dan: Came in by merge as far as i know, but yes i will remove it

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
Oran-Dan: Yes true

identity-wallet/src/state/credentials/reducers/self_issue_credential.rs
Oran-Dan: Yes, we don't follow any standard for our self-issued credentials and we don't have to, since we are planning to keep the self-issuance simple and compile the PID SD-JWT on the fly from them

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
Oran-Dan: Yeah that's more diligent

identity-wallet/src/state/credentials/reducers/send_token_request.rs
Oran-Dan: yes I like it

identity-wallet/src/state/credentials/reducers/send_token_request.rs
Oran-Dan: yes indeed it much needed a refactor, done

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
Oran-Dan: Well this error is kicked "upstream" and there the error is handled by the command handler as action failed so keep old state. I should write that more specifically maybe

identity-wallet/src/state/dev_mode/reducers/ferris_static_profile.rs
```diff
@@ -307,6 +307,8 @@ pub async fn load_ferris_profile() -> Result<AppState, AppError> {
     state.show_dev_mode_setting = true;
     state.dev_mode = DevMode::OnWithAutologin;
 
+    warn!("{state:#?}");
```
daniel-mader: Why log the `state` on warn level?

daniel-mader (dismissed):

daniel-mader (dismissed):

identity-wallet/src/state/credentials/reducers/refresh_all_credential_statuses.rs
```diff
@@ -0,0 +1,46 @@
...
+    for credential_id in credential_ids {
+        state = refresh_credential_status(state, Arc::new(RefreshCredentialStatus { credential_id })).await?;
+    }
+
+    // This line is needed to pass the correct `current_user_prompt` to the new state.
+    // The `current_user_prompt` differs depending on where this reducer will be used (currently only in UnlockStorage).
+    // So UnlockStorage needs the redirect prompt, most likely future use of this reducer in other locations won't.
+    let redirect_prompt = listen::<UnlockStorage>(action).map(|_unlock_storage| CurrentUserPrompt::Redirect {
+        target: "me".to_string(),
+    });
```
nanderstabel: The `UnlockStorage` action triggers these three reducers (in this order):
```rust
        vec![
            reducer!(unlock_storage),
            reducer!(sort_credentials),
            reducer!(refresh_all_credential_statuses),
        ]
    }
}
```

and the `unlock_storage` reducer initially sets the `current_user_prompt`, so the `refresh_all_credential_statuses` _should_ not have to reset it again. However, since `refresh_all_credential_statuses` calls `refresh_credential_status`, which unsets the `current_user_prompt` again, you had to reset it here. So I would suggest to change the comment to a `TODO` comment which states that this is a temporary solution and that we should solve it by immediately setting `current_user_prompt` to `None` when `main_exec` is called, before any reducer is called.

identity-wallet/src/state/credentials/reducers/refresh_credential_status.rs
Oran-Dan: you are right, i removed the error

identity-wallet/src/state/dev_mode/reducers/ferris_static_profile.rs
Oran-Dan: that's definitely not meant to be in there still, removed it

daniel-mader (dismissed):

nanderstabel (changes requested):

daniel-mader (approved):

nanderstabel (approved):

PR #593, opened 2025-08-21
feat: support deep linking for `openid://` and `openid-credential-offer://`

nanderstabel:
# Description of change
This pull request introduces deep linking functionality to the application, allowing it to be opened via custom URL schemes like `openid://` and `openid-credential-offer://`.

Key Features & Fixes
* Deep Link Integration:
  * The application now handles custom URL schemes, whether it's launched by the link or the link is received while the app is already running.
  * On desktop, this uses the single-instance plugin to ensure links are correctly forwarded to the existing application instance.
  * A `UniMe.desktop` file has been added to register the custom schemes on Linux (for testing/dev purposes).
* Race Condition Resolution:
  * A new `is_unlocked` flag has been added to the global `AppState`.
  * The frontend queues incoming deep links and only processes them once the `is_unlocked` flag is `true`, preventing errors that occurred when trying to process links before the wallet was unlocked.
* Performance Refactoring:
  * Fix: The `validate_jwt_vc_json` function has been refactored to accept the application's existing DID Resolver. This ensures that in Android the required `tls_config` is used. It also solves the performance cost of creating a new resolver instance on every credential validation.

n/a

Manually tested against ssi-agent `beta.11`

### DIFF
--- a/unime/UniMe.desktop
+++ b/unime/UniMe.desktop
@@ -1,5 +1,4 @@
 [Desktop Entry]
-Categories=
 Comment=Identity Wallet for people to manage Decentralized Identities and Verifiable Credentials
 Exec=gnome-terminal -- /usr/bin/unime %U
 Icon=unime
--- a/unime/UniMe.desktop
+++ b/unime/UniMe.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Comment=Identity Wallet for people to manage Decentralized Identities and Verifiable Credentials
-Exec=gnome-terminal -- /usr/bin/unime %U
+Exec=/usr/bin/unime %U
 Icon=unime
 Name=UniMe
 Terminal=false
--- a/unime/UniMe.desktop
+++ b/unime/UniMe.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Categories=
+Comment=Identity Wallet for people to manage Decentralized Identities and Verifiable Credentials
+Exec=gnome-terminal -- /usr/bin/unime %U
+Icon=unime
+Name=UniMe
+Terminal=false
+Type=Application
+MimeType=x-scheme-handler/openid;x-scheme-handler/openid-credential-offer
--- a/unime/src-tauri/Cargo.toml
+++ b/unime/src-tauri/Cargo.toml
@@ -29,11 +29,16 @@ rustls.workspace = true
 rustls-platform-verifier.workspace = true
 serde_json.workspace = true
 tauri-plugin-barcode-scanner = { version = "2" }
+tauri-plugin-deep-link = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v2" }
 tauri-plugin-fs = { version = "2" }
 tauri-plugin-keystore = { version = "2.1.0-alpha.1" }
 tauri-plugin-log = { version = "2", features = ["colored"] }
 tauri-plugin-os = { version = "2" }
 tauri-plugin-shell = { version = "2" }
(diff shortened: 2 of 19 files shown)


unime/UniMe.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Categories=
+Comment=Identity Wallet for people to manage Decentralized Identities and Verifiable Credentials
+Exec=gnome-terminal -- /usr/bin/unime %U
```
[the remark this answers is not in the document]
nanderstabel: This is done on purpose to make debugging easier on Linux. It indeed opens both UniMe and a terminal window where al the logs are displayed.

unime/src-tauri/Cargo.toml
```diff
@@ -29,11 +29,16 @@ rustls.workspace = true
 rustls-platform-verifier.workspace = true
 serde_json.workspace = true
 tauri-plugin-barcode-scanner = { version = "2" }
+tauri-plugin-deep-link = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v2" }
```
daniel-mader: Why not just `{ version = "2" }` like the other declarations?

unime/src-tauri/capabilities/main.json
```diff
@@ -4,6 +4,7 @@
   "windows": ["main"],
   "permissions": [
     "core:default",
+    "deep-link:allow-get-current",
```
daniel-mader: We can use `default` here, since it also only contains `allow-get-current` ([see here](https://tauri.app/plugin/deep-linking/#default-permission)).

unime/src-tauri/Cargo.toml
nanderstabel: because the update we needed was not published to crates.io yet when I made this, but it looks like they have now so I'll update it :)

unime/src-tauri/capabilities/main.json
nanderstabel: Will do :+1:

unime/UniMe.desktop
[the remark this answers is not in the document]
daniel-mader: Running this exec only works on Linux machines which use the GNOME desktop environment. Do you think we should handle this or just ignore it?

unime/UniMe.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Categories=
```
daniel-mader: Can be removed as it is optional.

unime/src-tauri/src/lib.rs
```diff
@@ -15,8 +15,18 @@ use jni::{
...
+    #[cfg(desktop)]
+    {
+        builder = builder.plugin(tauri_plugin_single_instance::init(|_app, argv, _cwd| {
+            info!("New app instance opened via deep link: {argv:?}");
+        }));
+    }
+
+    builder
         .invoke_handler(tauri::generate_handler![tauri_command::handle_action])
+        .plugin(tauri_plugin_deep_link::init())
```
daniel-mader: It should be safe to move this line to the other `.plugin()` functions below, right?

nanderstabel: Yeah should be safe, if you prefer you can also move:
```rust
    #[cfg(desktop)]
    {
        builder = builder.plugin(tauri_plugin_single_instance::init(|_app, argv, _cwd| {
            info!("New instance opened via deep link: {argv:?}");
        }));
    }
```

down to the other plugins. For `tauri_plugin_single_instance` it is important though to make sure it is initialized before all other plugins.

daniel-mader (approved):

PR #594, opened 2025-08-25
feat: support PIN code during issuance flow 

coplat:
# Description of change
adds support for the use of `tx_code` when specified as required in the Credential Offer. 
The inputted tx_code (for the user this is known as the PIN code) is then sent within the Token Request to the authorization server and validated. 

**BLOCKED BY**
- #596 

Be sure to reference any related issues by adding `fixes issue #`.

- [ ] I have followed the contribution guidelines for this project


identity-wallet/src/state/credentials/reducers/send_credential_request.rs
```diff
@@ -25,10 +25,9 @@ use uuid::Uuid;
 // confusing git diffs.
 pub async fn send_credential_request(state: AppState, action: Action) -> Result<AppState, AppError> {
     info!("send_credential_request");
+    if let Some(selected_offer) = listen::<CredentialOffersSelected>(action.clone()) {
```
nanderstabel: nit: Let's keep the empty line after the log message :)

identity-wallet/src/error.rs
```diff
@@ -100,6 +100,8 @@ pub enum AppError {
     TrustListNotFoundError(String),
     #[error("Failed to migrate AppState version `{0}` to version `{1}`: {2}")]
     AppStateMigrationError(u32, u32, String),
+    #[error("Provided code is invalid")]
+    InvalidTransactionCode(#[source] anyhow::Error),
```
nanderstabel: This error is not used anywhere so let's leave it out of this PR. Once we actually implement proper error handling for this we can add it back in but I suspect we would not want to return an error, but instead prompt the user to try to enter their pincode again (via `CurrentUserPrompt`)

unime/src/routes/prompt/credential-offer/+page.svelte
```diff
@@ -25,8 +27,20 @@
 
   let loading = false;
 
+  $: complete = false;
+
   const imageId = logo_uri ? hash(logo_uri) : '_';
 
+  const pinInput = new PinInput({
+    type: tx_code?.input_mode ?? 'numeric',
+    maxLength: tx_code?.length ?? 4,
```
nanderstabel: Does this mean that if no length is provided that by default there will be 4 slots? If so, would that mean that if the pincode is 6 digits, but the issuer did not provide the `length` property that the User won't be able to enter their full pincode?

identity-wallet/src/state/credentials/reducers/send_credential_request.rs
```diff
@@ -116,10 +115,25 @@ pub async fn send_credential_request(state: AppState, action: Action) -> Result<
...
+
+                    if tx_code_required && tx_code.is_none() {
+                        return Err(AppError::Error("tx_code is required but not provided".to_string()));
+                    }
+
+                    if tx_code_required {
+                        info!("tx_code is required and provided: {}", tx_code.is_some());
+                    } else {
+                        info!("tx_code not required for this offer");
+                    }
```
nanderstabel: imo I don't think we need these log messages here, but I think it would be nice to have a log message under `let tx_code_required = pre_authorized_code.tx_code.is_some();` stating something simple like:

```rust
                    info!(
                        "Handling Pre-Authorized code grant. Transaction code required: `{tx_code_required}`"
                    );
```

nanderstabel (changes requested):

unime/src/routes/prompt/credential-offer/+page.svelte
daniel-mader: Good catch! We could also just default to an simple text input with unlimited length. That would look a bit messy, but it would catch that mismatch. What scenario could you picture where the issuer doesn't set the length of the expected input?

nanderstabel: It's optional so unfortunately we cannot assume all Issuers will make use of it. I can not really think of a reason why anyone would explicitly not want to set it (maybe when for some reason the transaction code is based on a security question and therefore providing `length` would be giving somewhat of a clue to malicious actors?)

coplat: Could we not just default this to 6? Since that is the default tx_code length if the length property is not used?

coplat: Sounds like defaulting to a simple text input might be the easiest option. Although I can't imagine anyone needing a pincode of over 6 digits given current conventions - so that could be the secret max length?

daniel-mader: I agree, it's unlikely there will ever be more than 6 digits. However, we need to fallback to "unlimited" if it's not explicitly provided. We can't just render 6 fields by default if nothing is provided, because it could also be less than that. I don't really understand what's the reasoning behind making that field optional since every UI now has that same question.

daniel-mader (approved):

PR #599, opened 2025-08-30
feat: display SD-JWT secured W3C VC Credentials

nanderstabel:
# Description of change

This PR introduces support for `vc+sd-jwt` and `dc+sd-jwt` credential formats, enabling the wallet to correctly parse, store, display, and present them (presenting `vc+sd-jwt` credentials is still `TODO`).

#### 1. Dependency Fixes & Updates
-   **`sd-jwt-payload` Patch**: Patched the `sd-jwt-payload` crate to a fork that adds the necessary `Send + Sync` bounds to the `Hasher` trait. This resolves a critical build error that prevented the use of SD-JWT builders within our multi-threaded `async` architecture (both in Tauri commands and the state reducer system).
-   **`identity.rs` Patch**: Temporarily patched the `identity.rs` suite of crates to a fork. This was necessary to introduce changes that allow for distinguishing between `vc+sd-jwt` and `dc+sd-jwt` formats and to enable the creation of key-binding JWTs for presentations.

#### 2. Core Credential Handling Refactor
-   **Issuer-Defined Display Claims**: The `VerifiableCredentialRecord` logic has been overhauled. It now uses a new `try_new` constructor that accepts `ClaimDescription` metadata from the issuer.
-   **`DisplayClaim` Struct**: Introduced a new `DisplayClaim` struct that stores a claim's path, its issuer-defined display name (`key`), and its `value`.
-   **Enhanced `DisplayCredential`**: The `DisplayCredential` struct now includes a `display_claims` field. This allows the UI to render claims using the names and order intended by the issuer, rather than just iterating over raw JSON keys.
-   **Format Distinctions**: The parsing logic now correctly identifies `vc+sd-jwt` vs. `dc+sd-jwt` types based on the JWT `typ` header, storing the correct format in the credential record.
- 
**Warning:** presenting SD-JWT secured W3C VC Credentials is NOT supported yet

#### 3. SD-JWT Presentation
-   **`handle_oid4vp_authorization_request`**: Implemented the presentation flow for `dc+sd-jwt` credentials. This includes:
    -   Parsing the stored SD-JWT.
    -   Generating a `key_binding_jwt` signed by the subject's DID.
    -   Attaching the key-binding JWT to create a final, presentable SD-JWT string.

#### 4. UI Enhancements
-   **`DefaultRenderer.svelte`**: The credential renderer component has been updated to be aware of the new `display_claims` field.
    -   If a credential is an SD-JWT and has `display_claims`, it will now render the claims based on this structured data.
    -   For all other credential formats, it gracefully falls back to the previous method of rendering the raw `credentialSubject`.

n/a

Manually tested against the ilabs DIIPv4 conformance test suite.


identity-wallet/src/state/credentials/mod.rs
```diff
@@ -3,28 +3,47 @@ pub mod reducers;
...
+
 /// A credential displayable by the frontend.
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, Derivative, TS, Default)]
 #[derivative(PartialEq)]
 #[ts(export, export_to = "bindings/credentials/DisplayCredential.ts")]
 pub struct DisplayCredential {
     #[derivative(PartialEq = "ignore")]
     pub id: String,
-    #[ts(type = "string")]
+    #[ts(type = "{ format: string }")]
```
daniel-mader, line 38: This results in frontend code like this:

```
{
  format: {
    format: 'dc+sd-jwt'
  }
}
```

Is that intended?

nanderstabel: Yes. All `CredentialFormats` variants are essentially serialized as objects that always contain a `format` field and conditionally can have extra fields (depending on the concrete type of `C` in:
```rust
pub enum CredentialFormats<C = ()>
where
    C: FormatExtension,
...
```

So, usually we use it in combination with `#[serde(flatten)]` so that the nested fields are added to the root object (in this case `DisplayCredential`) during serialization, e.g.:
```rust
    #[serde(flatten)]
    pub format: CredentialFormats,
```

We cannot do that here unfortunately because the `TS` trait does not work with `#[serde(flatten)]`.

daniel-mader: Okay, thanks for clarifying 👍

daniel-mader (approved):

PR #614, opened 2025-09-03
feat: support use of external authorization servers 

### LINKED ISSUE #602: Allow use of external Authorization Servers
### Description

Currently, UniMe makes the assumption that the issuer is also the authorization server. For flexibility, wallets should be able to make use of external authorization servers which may be pre-defined by the issuer in their `CredentialIssuerMetadata` in the [`authorization_servers`](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-p) array. 

When this parameter is present, we can make use of one of listed authorization_servers in this array. If the parameter is empty, we assume that the authorization_server_url is the same as the credential_issuer_url. 

This is also highly relevant to us as we become compliant with the latest[ DIIP interoperability profile (v4),](https://fidescommunity.github.io/DIIP/#oid4vci) where wallets should **not** make the assumption  that the Authorization Server is on the same domain as the Issuer. 

Since the `authorization_servers` array can contain several authorization servers, we can select from them based on grant_type (as both pre-authorized and authorized flows also have an authorization_server parameter). 
If none are specified, we should select the first authorization_server in the array, although further flexibility could be later considered from a UX perspective (i.e. allowing the client/user to pick which authorization server they wish to use).  

### Resources

[OID4VCI Credential Issuer Metadata](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-p) 
[DIIP V4 Interoperability Profile](https://fidescommunity.github.io/DIIP/#oid4vci)

coplat:
# Description of change
Up till now, we have made the assumption that the issuer acts also the authorization server. Wallets should be able to make use of external authorization servers. 

Both grant types can optionally specify an authorization server of choice in the `authorization_server` parameter. 
If one is specified, the wallet must make use of it, though it is also necessary to validate that this authorization server is also listed in the `authorization_servers` array in the Credential Issuer Metadata. If it isn't, the fallback assumes the `authorization_server_url` is the `credential_issuer_url`. 

If the grant types do not specify an authorization server but the Credential Issuer Metadata's `authorization_servers` array contains one/some, then the first in the array is selected. 

If no authorization servers are specified, then we make the assumption that the issuer is the authorization_server as before. 

fixes issue #602 


identity-wallet/src/state/credentials/reducers/send_token_request.rs
```diff
@@ -79,9 +81,85 @@ pub async fn send_token_request(state: AppState, action: Action) -> Result<AppSt
...
+                .and_then(|grants| grants.authorization_code.as_ref())
+                .and_then(|auth_code| auth_code.authorization_server.as_ref())
+                .cloned()
+                .or_else(|| {
+                    if !credential_issuer_metadata.authorization_servers.is_empty() {
+                        Some(credential_issuer_metadata.authorization_servers[0].clone())
+                    } else {
+                        None
+                    }
+                })
```
nanderstabel: You can simplify this to:
```rust
.or_else(|| credential_issuer_metadata.authorization_servers.first().cloned())
```

Perhaps also add a `TODO` comment on top of it that states that we should in the future enable the User to select their preferred Authorization Server.

identity-wallet/src/state/credentials/reducers/send_token_request.rs
```diff
@@ -79,9 +81,85 @@ pub async fn send_token_request(state: AppState, action: Action) -> Result<AppSt
...
+                // Fall back to credential issuer url if no authorization server is specified.
+                .unwrap_or_else(|| credential_issuer_url.clone())
+        };
```
nanderstabel: The `if` and `else` blocks are nearly identical; only the grants differ, so see if you can 'move' the `if is_pre_authorized { .. } else { .. }` logic inside an `and_then` to keep the duplicated code at a minimum. :)

identity-wallet/src/state/credentials/reducers/send_token_request.rs
```diff
@@ -79,9 +81,85 @@ pub async fn send_token_request(state: AppState, action: Action) -> Result<AppSt
...
+                            .authorization_servers
+                            .contains(specified_auth_server)
+                    {
+                        return Err(AppError::Error(format!(
+                                "The specified authorization server {specified_auth_server} is not listed in the credential issuer metadata."
+                            )));
+                    }
+                }
+            }
+        }
```
nanderstabel: the `if` and `else` blocks are nearly identical so I think you should try to keep the duplicated code at a minimum :)

identity-wallet/src/state/credentials/reducers/send_credential_request.rs
```diff
@@ -142,6 +145,36 @@ pub async fn send_credential_request(state: AppState, action: Action) -> Result<
...
+                        .and_then(|grants| grants.authorization_code.as_ref())
+                        .and_then(|auth_code| auth_code.authorization_server.as_ref())
+                        .cloned()
+                        .or_else(|| {
+                            if !credential_issuer_metadata.authorization_servers.is_empty() {
+                                Some(credential_issuer_metadata.authorization_servers[0].clone())
+                            } else {
+                                None
+                            }
+                        })
```
nanderstabel: See https://github.com/impierce/identity-wallet/pull/614/files#r2387310167

nanderstabel (changes requested):

nanderstabel (dismissed):

nanderstabel (approved):

PR #624, opened 2025-09-19
feat: update self-issued credentials content, add address flow

SoulaAndrikopoulos:
# Description of change
Updated the 'self-issued' credentials onboarding text and translated it across all supported languages. Additionally, a 'Address' onboarding page was added to support the texts.

Fixes issue: 'Update Self-Issued Credential Onboarding Text'.

Self-reviewed the code, tested the app, and made sure it passes tests.


identity-wallet/src/state/actions.rs
```diff
@@ -60,7 +60,7 @@ pub fn listen<T: ActionTrait + Clone>(action: Action) -> Option<T> {
 /// ```
 
 // TODO: remove this once we have a better way to export the TS types.
-mod bindings {
+pub mod bindings {
```
nanderstabel: I changed my mind on how to fix the Clippy lint warning. Instead of adding this `pub` I think it is better to add `#[allow(dead_code)]` to the `Action` enum (and remove the `pub` there as well), like this:
```rust

    #[allow(dead_code)] // <-- add this line
    #[derive(Serialize, Deserialize, TS)]
    #[serde(tag = "type")]
    #[ts(export, export_to = "bindings/actions/Action.ts")]
    enum Action { // <-- `pub` is removed here
        #[serde(rename = "[App] Get state")]
        GetState,
        ...
```

nanderstabel (changes requested):
(I've only reviewed:
* `identity-wallet/src/state/actions.rs`
* `unime/src/i18n/nl-NL/index.ts`)

Dutch translations look good to me :) 

Other than that, I'd suggest to update the title of this PR so that it adheres to the Angular Conventional Commit header format (see https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md#commit-header).
In short that means you would want to prefix your PR title with `feat: ` instead of `Feat/`.

identity-wallet/src/state/actions.rs
SoulaAndrikopoulos: I've added the changes :>

nanderstabel (dismissed):

unime/src/i18n/en/index.ts
```diff
@@ -424,6 +424,26 @@ const en = {
     ADDRESS: {
       TITLE: 'Address',
       DESCRIPTION: 'Add your home or work address',
+      INFO: {
+        NAVBAR_TITLE: 'Verified address',
```
daniel-mader: This is misleading, since the address is _not_ verified, but self-attested.

daniel-mader (approved):

PR #688, opened 2025-12-19
feat: add ELM  validation

Oran-Dan:
# Description of change
I've added support for ELM validation in the `validate_credential_types()` function.

To avoid confusion: the European Publication Office has converted the VC DM 1.1 into a JSON Schema as well but enhanced it to match the EDC with fields such as `validFrom` and `issued`, which are not in the original VC DM 1.1. 
Therefore, we now have 2 distinct VC DM 1.1 JSON Schemas. One is created by ourselves, true to the spec. The other is provided by European Publication Office as it is referenced in their EDC JSON Schema.

The old tests pass.
I've added a unit test for a valid ELM validation.
Run `cargo test --all` in the repo root to test.


identity-wallet/src/state/core_utils/helpers.rs
```diff
@@ -212,6 +223,12 @@ impl CredentialType {
                     _ => Err(AppError::InvalidCredentialFormatError),
                 }
             }
+            CredentialType::EuropeanDigitalCredential => {
+                // The current provided ELM EDC schema contains no specific context value, only the context value of the VC DM 1.1 it builds upon.
+                // Therefore, there is no way to determine the version except for the description.
+                // For now we will shortcut this as ELM schemas are still in development and only time wil tell the best way to determine versions once multiple schemas are published.
```
coplat: nit, typo: "and only time _will_ tell"

identity-wallet/src/jsonschemas/EuropeanDigitalCredentialV3_3.json
```diff
@@ -0,0 +1,3188 @@
...
+          "$ref": "#/$defs/Many!LangStringType"
+        },
+        "verificationStatus": {
+          "$ref": "#/$defs/ConceptType"
+        },
+        "elmSubject": {
+          "$ref": "#/$defs/EuropeanDigitalCredentialType"
+        }
+      },
+      "required": ["verificationStatus", "subject", "dcType"]
```
coplat, line 2382: "subject" is a required field here but doesn't have a corresponding property definition. Is `elmSubject` meant maybe?

Oran-Dan: that's a good question, I will point this out to the ELM-EDC maintainers. This is their responsibility and since it compiles and works already it shouldnt block this PR. I will update it when they've updated or explained it to me.

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
nanderstabel, line 1: There are still some leftover println! statements in this file :)

identity-wallet/src/state/core_utils/helpers.rs
```diff
@@ -119,34 +120,42 @@ pub fn validate_credential_against_schema(json_schema_path: String, data: &Value
...
+            .map_err(|_| {
+                AppError::Error(format!(
+                    "Failed to compile JSON Schema from serde_json::Value: {json_schema}"
+                ))
+            })?,
     };
 
     let errors: Vec<ValidationError> = schema.iter_errors(data).collect();
     if !errors.is_empty() {
+        println!("Validation errors: {errors:#?}");
```
nanderstabel: Can this be removed? Or should this be `error!` instead?

identity-wallet/src/state/core_utils/helpers.rs
```diff
@@ -316,16 +349,89 @@ mod tests {
...
+                "id": "did:example:abcdef1234567890",
+                "type": "Person",
+                "fullName": { "en": "John Doe" }
+            }
+        });
+    }
+
+    #[test]
+    fn credential_schema_validation_elm_edc_ok() {
+        let result = validate_credential_types(&EXAMPLE_BASIC_ELM_EDC);
```
nanderstabel: Here the value that is being passed to `validate_credential_types` is a `Value::Object`, however the code introduced [here](https://github.com/impierce/identity-wallet/pull/497/changes#diff-2e69963cbef7874e07fcea526bcbcabc5371579d773a3ae3fac63b2820283340R223) passes a `Value::String` which will then always result in `debug!("No credential type found, skipping validation");`.

identity-wallet/src/state/core_utils/helpers.rs
```diff
@@ -119,34 +120,42 @@ pub fn validate_credential_against_schema(json_schema_path: String, data: &Value
     let json_schema: Value = serde_json::from_reader(json_schema_file)
         .map_err(|_| AppError::Error("Failed to convert JSON Schema &str to serde_json::Value".to_string()))?;
 
-    // Select correct draft version for JSON Schema Validator
+    // Define the relative path to our jsonschema folder needed for the LocalRetriever
+    let jsonschema_dir = std::env::current_dir().unwrap().join("resources/jsonschemas");
```
nanderstabel: This will panic on mobile (or in any environment where the application runs without the source code) because `resources/*` is not part of the binary. I think there are at least two ways to deal with this:
1. Take a look at `initialize_storage` and how we make use of Tauri's `data_dir` method.
2. Easier option (and therefore probably preferred): since we're dealing with static data here (in contrast to `STATE_FILE`, `STRONGHOLD`, etc.. --> Embed the JSON Values directly in to the code the same way we do for Ferris's test Credentials in `ferris_static_profile.rs`.

nanderstabel (changes requested):

identity-wallet/src/state/core_utils/helpers.rs
nanderstabel: If you go with option 2 I think you should still be able to keep your json files the way they are if you just 'include' them into the code through `include_str!`.

identity-wallet/src/state/core_utils/helpers.rs
Oran-Dan: fixed

identity-wallet/src/state/did/validate_linked_verifiable_presentations.rs
Oran-Dan: fixed

identity-wallet/src/state/core_utils/helpers.rs
Oran-Dan: fixed

identity-wallet/src/state/core_utils/helpers.rs
Oran-Dan: took a slightly different approach but i think we are still aligned

identity-wallet/src/state/core_utils/helpers.rs
```diff
@@ -70,172 +64,53 @@ pub async fn get_issuer_document(resolver: &Resolver, credential_jwt: &Jwt) -> O
...
-#[derive(serde::Deserialize, PartialEq, Debug, strum::Display)]
-enum CredentialTypeVersion {
-    VerifiableCredentialV1_1,
-    VerifiableCredentialV2,
-    OpenBadgeCredentialV3,
-    #[serde(other)]
-    Unknown,
+/// This trait is solely to add a method to serde_json::Value for converting Values to Strings cleanly
+pub trait ValueToString {
+    fn to_clean_string(&self) -> Option<String>;
```
daniel-mader, line 105: Could you add a few unit tests for this function (happy cases and also error cases)? This could help with documenting what this function does and why we need it. I do understand what you mean by "clean string", but maybe we can use a more common term such as "unescaped" or "unquoted" or similar.

PR #692, opened 2026-02-03
fix: separate parsing logic of VCDM2.0 and SD-JWT VC Credentials

nanderstabel:
# Description of change
This change separates the parsing of VCDM2 SD-JWTs (`vc+sd-jwt`) from SD-JWT VCs (`dc+sd-jwt`), and prevents the error below (the `vct` claim is REQUIRED in SD-JWT VC but not for SD-JWT secured VCDM2 Credentials).

```rust
#[test]
fn test() {
    let test = "eyJ0eXAiOiJ2YytzZC1qd3QiLCJraWQiOiJkaWQ6d2ViOmxvY2FsaG9zdCUzQTMwMzMjQXM1cjRVeG9fWjREV19XLW9BUVNURTRLT1hhMnVWeUJRYUNna3VTUkhwWSIsImFsZyI6IkVTMjU2In0.eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiXSwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCJdLCJjcmVkZW50aWFsU3ViamVjdCI6eyJfc2QiOlsiVk5BZUVsM0d1YnVrQWx0TVNRUEhJZ25qcm9XckZ6eDd5NW5lQkgxV19vOCIsIl9UZ3owMmhiYmJKbzBUM2NFV3BRbkh4WjNxZGdLcDZyd2w1Ri1PeW5NN0kiLCJ5MnhKZ01NQjJOajVqQ09idXpBWVNiaXZYWncwRmg4VmN2UzlzeTlrQUZnIl19LCJpc3N1ZXIiOnsiaWQiOiJodHRwOi8vbG9jYWxob3N0OjMwMzMvIiwibmFtZSI6IlVuaUNvcmUifSwidmFsaWRGcm9tIjoiMjAyNi0wMi0wM1QyMDo0OTo1MloiLCJjcmVkZW50aWFsU3RhdHVzIjp7ImlkIjoiaHR0cDovL2xvY2FsaG9zdDozMDMzL2lldGYtb2F1dGgtdG9rZW4tc3RhdHVzLWxpc3QvMCIsInR5cGUiOiJzdGF0dXNsaXN0K2p3dCIsImlkeCI6Mjg1OCwidXJpIjoiaHR0cDovL2xvY2FsaG9zdDozMDMzL2lldGYtb2F1dGgtdG9rZW4tc3RhdHVzLWxpc3QvMCJ9LCJzdGF0dXMiOnsic3RhdHVzX2xpc3QiOnsidXJpIjoiaHR0cDovL2xvY2FsaG9zdDozMDMzL2lldGYtb2F1dGgtdG9rZW4tc3RhdHVzLWxpc3QvMCIsImlkeCI6Mjg1OH19LCJfc2RfYWxnIjoic2hhLTI1NiIsImNuZiI6eyJraWQiOiJkaWQ6andrOmV5SmhiR2NpT2lKRlV6STFOaUlzSW1OeWRpSTZJbEF0TWpVMklpd2lhMmxrSWpvaU9VaDFObEZETFdzM0xWVjZZMDF3TWpGT1Z6ZDFWRzU0V0VsMVEyZHNaRkpMYUVGWE0xaEdTSE5mVlNJc0ltdDBlU0k2SWtWRElpd2llQ0k2SW1ab1ZWWjNTbGd0U2tKNUxVazVWa1JVTUdSNlpXVlZkVXRSV1dSa2JFbExhRVl6VkROUlozZEVlVmtpTENKNUlqb2lUMUk0VVVsWFpVVmtjRTlwU0haUVYyWkZPWGxKVEdJM1RFcDNXVE01UVRCelJqZEZSemhqV0ZZMVFTSjkjMCJ9fQ.1OKm-xew82vumRIhm8fQAoZkflViUokvQ1M2VMdWvHuI-RUQWMsABZEqnoDrOvAiuI8JkjPDeZVo0nDEg-8S5Q~WyJVcGFQWlV2SkMyMnE2NEVMNG4yWWFuTDA4amxTTGMxZ1pPLWNFNHR6IiwiZG9iIiwiMTk4Mi0wMS0wMSJd~WyJwRmo1S01jYURPaDBFQ0ZYSWRwZXR5OHF6cm9jZ1RoX0JQb05HUjFNIiwiZmlyc3RfbmFtZSIsIkZlcnJpcyJd~WyJaSGZMSnlGQU1XaE51UnJNNlgwbzR5X21fRnYwd3JjR3JIN1h1X0ZJIiwibGFzdF9uYW1lIiwiQ3JhYm1hbiJd~";

    let test = SdJwtVc::parse(test).unwrap();
}
// called `Result::unwrap()` on an `Err` value: MissingClaim("vct")
```

- https://github.com/impierce/ssi-agent/pull/278

- Updated the `test_verifiable_credential_record_try_from_sd_jwt_vc` test
- Added the `test_verifiable_credential_record_try_from_vcdm2_sd_jwt` test
- Validated that all test pass in the [ITB DIIPv5 Test Suite](https://itb.ilabs.ai/)


identity-wallet/src/state/credentials/mod.rs
```diff
@@ -171,6 +136,42 @@ impl VerifiableCredentialRecord {
                     "credentialSubject": credential_subject
                 });
 
+                (id, format, data, issuance_date, display_claims)
+                // Else
```
Oran-Dan: Can the // Else be removed?

identity-wallet/src/state/credentials/mod.rs
Oran-Dan: After checking the parse and FromStr implementations of `SdJwtVC` and `SdJwt` I don't see a type header check anywhere. So I think we still need to manually check the typ header as we do with `vc+sd-jwt`

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -108,58 +113,18 @@ impl VerifiableCredentialRecord {
             {
                 info!("Verifiable Credential parsed as a SD-JWT VC");
 
-                let typ = sd_jwt_vc.header().get("typ").and_then(|typ| typ.as_str());
-
                 let id = Uuid::new_v4().to_string();
                 let issuance_date = sd_jwt_vc.claims().iat.map(|iat| iat.to_rfc3339()).unwrap_or_default();
 
-                let mut credential_subject = serde_json::json!(sd_jwt_vc
+                let credential_subject = serde_json::json!(sd_jwt_vc
```
Oran-Dan: I think the misunderstanding that all relevant data lives in the credential_subject still lives here. ~A simple rename `credential_subject` -> `credential` would suffice. The heavy lifting has to happen before line 133 where the entire credential(_subject) is put into the `credential_subject` field. To figure this out properly is scope creep but also definitely tech-debt. I'll leave it up to you to decide whether you want to tackle that in this PR. Whatever you choose I still recommend renaming this variable.~

See comment below, there i suggest keeping the dc+sd-jwt as is and solving the tech-debt without complicating this any further, nor just leaving this incorrect code to live on.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -108,58 +113,18 @@ impl VerifiableCredentialRecord {
...
-                    .collect();
-
-                let format = if let Some("dc+sd-jwt") = typ {
-                    CredentialFormats::DcSdJwt(())
-                } else {
-                    CredentialFormats::VcSdJwt(())
-                };
+                let format = CredentialFormats::DcSdJwt(());
 
                 // TODO: Remove this workaround that is basically a way of disguising the SD JWT VC as a VC so that
```
Oran-Dan: I think we should solve this tech-debt first, also in relation to the TODO right below this and my previous review comment. Since the whole workaround is actually incorrect and complicates things a lot if we want to do it the right way. It is much easier to keep the dc+sd-jwt as is and make sure the frontend displays it correctly. If i'm not wrong this doesnt have to be too much work in the frontend anyway.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -171,6 +136,42 @@ impl VerifiableCredentialRecord {
...
+                let type_ = credential.types.clone();
+                let credential_subject = credential.credential_subject.first().unwrap().clone();
+
+                let display_claims: Vec<DisplayClaim> = get_display_claims(claim_descriptions, &json!(credential));
+
+                let format = CredentialFormats::VcSdJwt(());
+
+                let data = json!({
+                    "type": type_,
+                    "credentialSubject": credential_subject
```
Oran-Dan: here the conversion from vc+sd-jwt to a vc makes sense of course but still the `credentialSubject` misconception lives on. A lot of info in the root can be lost here. Keep in mind that additional properties are always allowed in the root and in actually any property in VC 1.1 and 2.0. This means if we want to keep it simple we can just copy the entire claims section into this `data` variable. `iss` claims and such won't break the schema validation.

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -171,6 +136,42 @@ impl VerifiableCredentialRecord {
...
+            } else if let Some(sd_jwt) = verifiable_credential
+                .as_str()
+                .and_then(|verifiable_credential| verifiable_credential.parse::<SdJwt>().ok())
+            {
+                let id = Uuid::new_v4().to_string();
+                let issuance_date = sd_jwt
+                    .claims()
+                    .get("validFrom")
+                    .and_then(|valid_from| valid_from.as_str())
+                    .unwrap_or_default()
```
Oran-Dan: This unwrap_or_default doesn't give us the desired behavior I think, the VC DM 2.0 makes it optional but then says something which actually implies more than they make it seem... 
```Note: Validity start period for Verifiable Credentials
If validFrom and validUntil are not present, the [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) validity period is considered valid indefinitely. 
In such cases, the [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) is assumed to be valid from the time the verifiable credential was created.
```
https://www.w3.org/TR/vc-data-model-2.0/#validity-period
This is close to implying that if an `iat` claim is also missing that the VC is actually still invalid. 
In my opinion we should enforce it as such. 
So first validFrom, if not present then iat, if that's also not present than the VC is invalid. 
What usecase can we think of where an iat/validFrom/issuanceDate is not an absolute necessity? 
And if we can think of one, does that edgecase outweigh the security risk of credentials with eternal validity into the past?

identity-wallet/src/state/credentials/mod.rs
Oran-Dan: Can we remove this hardcoded logic or not yet?

Oran-Dan: Since this is VC DM 1.1 this unwrap_or_default should also actually be a hard error: Invalid Credential

identity-wallet/src/state/credentials/mod.rs
```diff
@@ -252,6 +253,33 @@ impl VerifiableCredentialRecord {
     }
 }
 
+fn get_display_claims(claim_descriptions: Vec<ClaimDescription>, data: &serde_json::Value) -> Vec<DisplayClaim> {
+    claim_descriptions
+        .into_iter()
+        .map(|claim_description| {
+            let key = claim_description
+                .display
+                .first()
```
Oran-Dan, line 262: I think we should add a TODO here as well that this doesn't handle multiple values AKA multiple locales yet

Oran-Dan (changes requested):
I havent reviewed the tests yet as I would first like to discuss/address the current comments

identity-wallet/src/state/credentials/reducers/handle_oid4vp_authorization_request.rs
Oran-Dan: This line seems to be redundant

Oran-Dan: even though you didnt add this code, would be nice to clean up the following points (I also pointed this out to Cheyenne already):
- This fn doesnt need to be pub
- It's a helper and should therefore be below the "main" function.

identity-wallet/src/state/credentials/mod.rs
nanderstabel: I made some effort to make all of this more ergonomic but it is a deeper rabbithole than you'd think and we don't really have time to spend too much time on UniMe at this point unfortunately. Also, the frontend is already displaying the claims correctly.

I wish I could solve all of this properly (I made a good start in `refactor/verifiable-credential-record`) but for now I have to keep this PR scoped to just a fix (and not a refactor).

identity-wallet/src/state/credentials/mod.rs
nanderstabel: There is no mention of `iat`, I think it just means that if validFrom and validUntil are not present, the validity period is considered valid indefinitely. The fact that the credential exists means that it is created, no need to tie it to a specific claim imo.

identity-wallet/src/state/credentials/mod.rs
nanderstabel: removed

identity-wallet/src/state/credentials/reducers/send_token_request.rs
nanderstabel, line 1: Note: it looks as if there are a lot of changes in this file but that's because we had two `for` loops looping over the same data for no reason. I (had to) merge them into one single for loop so that `credential_configuration.credential_format.format()` could be supplied to the `VerifiableCredentialRecord::try_new` function at the end. The other change is related to commenting out `validate_jwt_vc_json`, reason for this can be read in the TODO comment.

Oran-Dan (dismissed):
Awesomesauce

daniel-mader (dismissed):

daniel-mader (approved):

PR #706, opened 2026-03-01
feat: move `Resolver` initialization to `Subject` (enabling offline startup)

Oran-Dan:
# Description of change
I have refactored the code around the Resolver struct to live only in 1 place and in a OnceCell so that it will only be initialized when actually used for the first time and not at startup.
This absolves the hang up time and enables offline startup.

I've included two minor refactors:
- Subject is not dynamic (`dyn`) anymore, as it did not need to be so and its always preferred to avoid dyn when possible (Better for memory management and keeps the code simpler)
- Resolver used to live in two places for no reason, in Subject and in CoreUtils. Now it only lives in Subject.

-

Since its really just a refactor which enabled the offline feature no new tests are needed. 


identity-wallet/src/state/common/reducers/unlock_storage.rs
```diff
@@ -34,7 +33,14 @@ pub async fn unlock_storage(state: AppState, action: Action) -> Result<AppState,
         )
         .map_err(OID4VCWalletError)?;
 
+        state_guard.identity_manager.replace(IdentityManager {
+            subject,
+            provider_manager,
+            wallet,
+        });
```
nanderstabel, line 40: Is there a specific reason why this needed to be moved up?

nanderstabel (dismissed):

nanderstabel (approved):

daniel-mader (approved):

